PR #3, opened 2018-09-30
Overwrite versionitis, use latest IIF code

dehann:
This PR is to replace old IIF code that was used to initially create DFG.  A few extra functions have been included for DFG, but not tested in this context.  Relates to packing and unpacking of factor and variable nodes.


GearsAD:
Cool, thanks!

PR #5, opened 2019-04-17
Proof of concept for API and Graphs.jl implementation

GearsAD:


src/entities/DistributedFactorGraph.jl
```diff
@@ -0,0 +1,22 @@
+import Base: show
+
+struct DistributedFactorGraph
+    # A cache - TBD
+    # something::T
+    # An API for communication
+    api::DFGAPI
+    # A series of mirrors that are asynchronously updated.
+    mirrors::Vector{DFGAPI}
```
dehann: Does this relate to the mirroring happens inside or outside DFG conversation?  I’m nervous about building too much automation into DFG, vs DFG sets the API, establishes symmetry, provides set operations, and lays the ground work for “operating system” features.  Also need to understand how much serialization and deserialization needs to occur in DFG, my guess is at least some would have to occur here and the rest through dispatch to used defined converters.  If you imply mirroring/duplication, I’m just concerned about too much.  Goes back to discussion that duplication happens above DFG in either RoME or Caesar?

src/services/GraphsDFG.jl
```diff
@@ -0,0 +1,432 @@
...
+    end
+    return true
+end
+
+"""
+    $(SIGNATURES)
+Get a DFGVariable from a DFG using its underlying integer ID.
+"""
+function getVariable(dfg::GraphsDFG, variableId::Int64)::DFGVariable
+    @warn "This may be slow, rather use by getVariable(dfg, label)"
```
dehann: Is this from test results?  My very brief experience was that Symbol lookups on a dict where much much slower than Int key lookups, and strong lookups were much slower than Symbol.  We should confirm.

src/services/GraphsDFG.jl
```diff
@@ -0,0 +1,432 @@
...
+
+"""
+    $(SIGNATURES)
+Update a complete DFGVariable in the DFG.
+"""
+function updateVariable!(dfg::GraphsDFG, variable::DFGVariable)::DFGVariable
+    if !haskey(dfg.labelDict, variable.label)
+        error("Variable label '$(variable.label)' does not exist in the factor graph")
+    end
+    dfg.g.vertices[dfg.labelDict[variable.label]].dfgNode = variable
```
dehann: I had cases where I ably want to update some of the fields, more a user thing and I’m not exactly sure of how the new use case will work, but just keep in mind till we have completed integration.

src/services/GraphsDFG.jl
```diff
@@ -0,0 +1,432 @@
...
+
+"""
+    $(SIGNATURES)
+Delete a DFGVariable from the DFG using its label.
+"""
+function deleteVariable!(dfg::GraphsDFG, label::Symbol)::DFGVariable
+    if !haskey(dfg.labelDict, label)
+        error("Variable label '$(label)' does not exist in the factor graph")
+    end
+    @error "Delete is not supported in Graphs.jl implementation at present"
```
dehann: I will add support for delete in graphs.jl

src/services/GraphsDFG.jl
```diff
@@ -0,0 +1,432 @@
...
+List the DFGFactors in the DFG.
+Optionally specify a label regular expression to retrieves a subset of the factors.
+"""
+getFactors(dfg::GraphsDFG, regexFilter::Union{Nothing, Regex}=nothing)::Vector{DFGFactor} = lsf(dfg, regexFilter)
+
+"""
+    $(SIGNATURES)
+Checks if the graph is fully connected, returns true if so.
+"""
+function isFullyConnected(dfg::GraphsDFG)::Bool
```
dehann: We should remember to add a ‘hasOrphans’ alias

src/entities/DistributedFactorGraph.jl
GearsAD: This stuff is older code, saving it for moving it out later. Right now it's inert code though

src/services/GraphsDFG.jl
GearsAD: The worry isn't actually on the lookup itself, rather on the check to see if the variable is there. Probably O(n) to check if it's in the values of the dictionary. If this is used, then it's probably a good idea to have a idDict
```
    if !(variableId in values(dfg.labelDict))
        error("Variable ID '$(variableId)' does not exist in the factor graph")
    end
```

src/services/GraphsDFG.jl
GearsAD: Yep! That's part of the actual DFGVariable itself though, more on that when adding the fields into the variables/factors.

src/services/GraphsDFG.jl
GearsAD: Done!

PR #19, opened 2019-06-01
few updates to Symbol

dehann:


dehann:
fixes JuliaRobotics/IncrementalInference.jl#256

PR #26, opened 2019-06-09
add JLD2 dependency, not working yet

dehann:
loading jld not working yet -- also do we even what this here?


dehann:
This might be a short lived idea:  Going for FileDFG instead: see #35

Issue #24, opened 2019-06-05
Tabs on previous ls sorting method with weird variable labels


dehann:
Just to keep a pointer on many hours spent on sorting method for variable labels:
https://github.com/JuliaRobotics/IncrementalInference.jl/blob/97b1ba88dfccf64bafb2bbe929b51b5647053d82/src/FGOSUtils.jl#L455

Will figure out how to re-purpose in new DFG.

PR #37, opened 2019-06-10
ls getVarIds now search tags too

dehann:


dehann:
still working on sorting the ls return list (currently affects plotting)

src/services/GraphsDFG.jl
```diff
@@ -342,6 +361,83 @@ Checks if the graph is not fully connected, returns true if it is not contiguous
...
+  while !allnums(st[i:i])  i+=1; end
+  return i
+end
+
+"""
+    $SIGNATURES
+
+Sort a variable list which may have nested structure such as `:x1_2` -- does not sort for alphabetic characters.
+"""
+function sortVarNested(vars::Vector{Symbol})::Vector{Symbol}
```
GearsAD: Is this used in GraphsDFG? If not, do you mind it to a common utils file?

dehann: think that should be better?

dehann:
solves #24

PR #41, opened 2019-06-16
CloudGraphsDFG implementation

GearsAD:


src/entities/DFGFactor.jl
```diff
@@ -1,3 +1,14 @@
+
+# Originally from IncrementalInference
```
GearsAD: @dehann Going to merge but not tag, this part will break IncrementalInference because of duplicate definitions.

dehann: I would like to tag as soon as possible even if it breaks IIF v0.7.0.  Let's plan on DFG v0.1.3 and IIF v0.7.1 in the next day or two.

cc @mc2922

PR #46, opened 2019-06-21
fix testing

dehann:


dehann:
rapid fix for requires based dependency on Neo4j and DataFrames

PR #57, opened 2019-07-07
starting migration of IIF low level functions to DFG, and fixes

dehann:


dehann:
xref JuliaRobotics/IncrementalInference.jl#342

PR #67, opened 2019-07-26
LightGraphs integration

GearsAD:


Affie:
Failures temporary allowed on v1.2 for Metagraphs. Should be fixed in master, waiting for new release.

Affie:
Looks like it should be MetaGraphs v0.6.4, but its not registered yet. v0.6.4 will hopefully fix build on julia 1.2

dehann:
FYI JuliaGraphs/MetaGraphs.jl#84

src/LightGraphsDFG/services/LightGraphsDFG.jl
```diff
@@ -0,0 +1,545 @@
...
+Get an adjacency matrix for the DFG, returned as a Matrix{Union{Nothing, Symbol}}.
+Rows are all factors, columns are all variables, and each cell contains either nothing or the symbol of the relating factor.
+The first row and first column are factor and variable headings respectively.
+"""
+function getAdjacencyMatrix(dfg::LightGraphsDFG)::Matrix{Union{Nothing, Symbol}}
+    varLabels = sort(map(v->v.label, getVariables(dfg)))
+    factLabels = sort(map(f->f.label, getFactors(dfg)))
+    vDict = Dict(varLabels .=> [1:length(varLabels)...].+1)
+
+    adjMat = Matrix{Union{Nothing, Symbol}}(nothing, length(factLabels)+1, length(varLabels)+1)
```
dehann: One thing we should start doing across the entire code base is avoid type instability -- a major performance hog.  It's okay to use Union for function dispatch, thats still fast, but dynamic types require run time type inference each time.  I think this is happening in various places in the current IIF code anyway, but likelihood is high that we will run a campaign to remove all type instability from the code to improve performance.

cc @GearsAD

dehann: I'm not sure if we will get type stability at the end of IIF v0.7.x (probably not) or rather do that cleanup during IIF v0.8.x cycle.  v0.8.x cannot be as unstable as IIF v0.7.x is though.  v0.7.x is bit of a hot mess.  As long as IIF 0.6.x and v0.8.x are reliable form version control standpoint.  v0.7.x is a little tricky in places.

Affie: Should we change to missing? is missing and nothing treated the same? As far as I know missing has little performance impact.

Affie:
> FYI [JuliaGraphs/MetaGraphs.jl#84](https://github.com/JuliaGraphs/MetaGraphs.jl/issues/84)

src/LightGraphsDFG/services/LightGraphsDFG.jl
dehann: I don't know missing that well unfortunately.  The union typedef (not function dispatch) is a bigger story.  This is part of the current DFG design so im scared if LightGraphsDFG is out of step then things are going to get weird.  My first reaction is to just update all of DFG in one go.

GearsAD: In this case we need a nullable type. TBD - Let's do DFG in one go?

Affie: I don't think we need to do anything in this case, julia arrays are fast with Nothing unions. 
Union{Nothing,T} = nullable type.

GearsAD:
LGTM!

GearsAD:
Created [IIF #368](https://github.com/JuliaRobotics/IncrementalInference.jl/issues/368) to capture effort  to support this driver for solving.

src/LightGraphsDFG/services/LightGraphsDFG.jl
dehann: FYI [avoid containers with abstract types](https://docs.julialang.org/en/v1/manual/performance-tips/#Avoid-containers-with-abstract-type-parameters-1)

PR #68, opened 2019-07-29
Option with removed nodeCounter and labelDict

Affie:
it is now passed to the MetaGraphs dict :label
nodeCounter just returs the number of nodes nv(G)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,6 @@ notifications:
 matrix:
   allow_failures:
     - julia: nightly
-    - julia: 1.2
 
 # script:
 #   - if [[ -a .git/shallow ]]; then git fetch --unshallow; fi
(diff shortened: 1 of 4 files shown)


Affie:
I don't know if nodeCounter or labelDict will be used externally. If not used I would suggest removing it from the structure and optionally passing it on to metagraphs (they did all the hard work already in keeping it up to date). It should be Read Only!

dehann:
HI thanks, yes want to remove that too but it has been kept for legacy transitions from IIF v0.6.x.  We can remove integer based labels and lookup for nodes in the graph once we have parity on API and good handle on performance.  The plan is to remove unless some showstopper comes up.  Also, it may be best to remove the nodeCounter and lookup dict from all DFG objects at the same time, so maybe we should point this PR for DFG v0.3.x cycle?

Affie:
Hi, so far I pointed it inward to use the metagraph dict. It should still be compatible as long as its only read. If it is used externally I would have to rethink the whole driver.

dehann:
Don't worry about the nodeCounter.  We dont want to use Int.  I think this is good.  We should then just sync with IIF v0.8.x release.  I think the driver is good as it is here!

PR #70, opened 2019-08-02
add codecov badge

Affie:
looks like codecov is used in the rest of caesar, is both needed?


dehann:
not sure if both are needed, but there is something wrong with the code coverage badge -- haven't dug into why yet.  I'l merge this and see if that fixes it, thanks!

Affie:
The other badge is for coveralls, and repo is not registered there. Do we want/need both of them?

PR #79, opened 2019-08-14
Sparse Int Adjacency matrix

Affie:
PR For discussion @GearsAD @dehann

Currently Dense `Union{Nothing, Symbol}`
```
@time getAdjacencyMatrix(dfg)
  0.000129 seconds (757 allocations: 87.281 KiB)
```

Working with spare matrix as Adjacency Matrix
 ```
@time exp_getAdjacencyMatrix(dfg)
  0.000013 seconds (11 allocations: 1.531 KiB)
```
```julia
23×23 SparseArrays.SparseMatrixCSC{Int64,Int64} with 48 stored entries:
  [11,  1]  =  1
  [11,  2]  =  1
  ⋮
  [9 , 23]  =  1
  [20, 23]  =  1
```
Navigation type factor graphs adjacency will mostly be very sparse 
1. Should sparse matrices (or maybe adjacency lists if symbols are used) be used throughout DFG getAdjacencyMatrix
2. How do we want to handle labels. 
     - Sparse internal id (`Int`) sparse matrix with a map for the `Int`s <-> labels. 
     - Sparse label matrix
     - List of adjacent labels 
     - Fastest most efficient would be to use iterators and `Int` in IIF. Although, it would only make sense with in memory types.

### DIFF
--- a/src/LightGraphsDFG/LightGraphsDFG.jl
+++ b/src/LightGraphsDFG/LightGraphsDFG.jl
@@ -16,7 +16,7 @@ export ls, lsf, getVariables, getFactors, getVariableIds, getFactorIds
 export getVariable, getFactor
 export updateVariable!, updateFactor!
 export deleteVariable!, deleteFactor!
-export getAdjacencyMatrix
+export getAdjacencyMatrix, getAdjacencyMatrixSparse
 export getAdjacencyMatrixDataFrame
 export getNeighbors
 export getSubgraphAroundNode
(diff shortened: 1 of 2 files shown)


dehann:
I like this sparse approach and dropping the Union type from data structure.  Think we should target DFG v0.4.x or this, given the breaking change.  What do you think @GearsAD ?

Affie:
I had a quick look at IIF and it can be modified as follows to still work without much changes:
```julia
adjMat, permuteds, permutedsf  = DFG.getAdjacencyMatrix(dfg)
A = Array(adjMat)
```

PR #88, opened 2019-09-02
Integrating @affie's changes into master (v0.4 release)

GearsAD:


src/services/interfaces.jl
```diff
@@ -0,0 +1,8 @@
+
+
+"""
+    $(SIGNATURES)
```
GearsAD: Like this! Can we put this in services/AbstractDFG.jl?

PR #125, opened 2019-09-28
draw graph circles squares

### LINKED ISSUE #56: [REQUEST]: swap squares and circles when drawing dot of dfg object
The SLAM community seem to prefer factors as squares and variables as circles.  At some point we should maybe just swap the picture.  Relating to graphviz visualization.

dehann:
closes #56


dehann:
thanks @tonioteran

dehann:
think there is a dot file export check that we need to update too.

Issue #97, opened 2019-09-05
Add big data


Affie:

Affie:
Store keys=>AbstractBigData vs. only keys
 
keys=>AbstractBigData
- To guarantee small payload: 
  - only dispatch to cloud on specific <:AbstractBigDataElement types. 
  - In memory (should) pass pointers around so it should ideally not matter.
- On delete variable: big data entry should be deleted externally

Affie:
WIP PR https://github.com/JuliaRobotics/DistributedFactorGraphs.jl/pull/131

GearsAD:
Updated in #228 using your AbstractBigDataPattern

GearsAD:
AbstractBigDataEntry looks good - added a `GeneralBigDataEntry`, should support general use-cases - https://github.com/JuliaRobotics/DistributedFactorGraphs.jl/pull/228/files#diff-a9bf58a6f5bf50be944b4eb28d78862eR6-R12

Affie:
Confirm if this is in CGDFG

PR #131, opened 2019-10-02
Add BigData

Affie:
Big data added to DFGVariable

### DIFF
--- a/src/entities/DFGVariable.jl
+++ b/src/entities/DFGVariable.jl
@@ -1,7 +1,11 @@
+#TODO don't know what to do if it is uninitalized
+#so for now defining a Singleton for the default
+struct SingletonInferenceVariable <: InferenceVariable end
+
 """
 $(TYPEDEF)
 """
-mutable struct VariableNodeData
+mutable struct VariableNodeData #TODO v0.5.0 {T<:InferenceVariable}
   val::Array{Float64,2}
   bw::Array{Float64,2}
   BayesNetOutVertIDs::Array{Symbol,1}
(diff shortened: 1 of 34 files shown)


Affie:
Initial idea for big data to get us going

src/entities/DFGVariable.jl
```diff
@@ -90,14 +91,22 @@ mutable struct DFGVariable <: AbstractDFGVariable
     estimateDict::Dict{Symbol, Dict{Symbol, VariableEstimate}}
     solverDataDict::Dict{Symbol, VariableNodeData}
     smallData::Dict{String, String}
-    bigData::Any
+    bigData::Dict{Symbol, AbstractBigDataEntry}
```
dehann: can we use Option 2 method here, i.e.:
```julia
bigData::Dict{Symbol, <: AbstractBigDataEntry}
```
https://github.com/JuliaRobotics/DistributedFactorGraphs.jl/pull/148#issuecomment-541536774

Affie: The whole advantage from this approach is to have an abstract container. I did some benchmarks and an abstract dict with composite types in does not seem to perform worse than a typed dict.
I'm more than open for suggestions of other ways to do it, but <: AbstractBigDataEntry will not work in this case.

The advantage is in the flexibility, the types I added for illustration and the ones I currently find useful are:
```juila
struct MongodbBigDataEntry <: AbstractBigDataEntry
    key::Symbol
    oid::NTuple{12, UInt8} #mongodb object id
    #maybe other fields such as:
    #flags::Bool ready, valid, locked, permissions
    #MIMEType::Symbol
end

struct FileBigDataEntry <: AbstractBigDataEntry
    key::Symbol
    filename::String
end
```
One can just as easily define other types as long as it has a key field, maybe in memory - stored in an dictionary:
If you don't mind it not working in the cloud.
```juila
struct MemoryBigDataEntry <: AbstractBigDataEntry
    key::Symbol
    dictionaryModule::Symbol
    dictionaryReference::Symbol
end
```

The alternatives i could think of were to standardize `BigDataEntry`, but I don't know what to put in there and you lose dispatch. Or to just keep the keys in a `Set{Symbol}` kind of way.

GearsAD:
We can close this right, it's merged into master already as part of megaPR?

Affie:
Looks like some of the tests got lost in the merge conflicts.

dehann:
xref #97

PR #140, opened 2019-10-07
[WIP] GraphsDFG and LightGraphsDFG implementation of user/robot/session data

GearsAD:


GearsAD:
This still needs the changes for `MetaGraphDFG` and `SymbolDFG`

GearsAD:
Still TODO:
- MetaGraphDFG and SymbolDFG
- FileDFG serialize/deserialize
- CloudGraphsDFG serialize/deserialize

src/LightDFG/services/LightDFG.jl
```diff
@@ -13,6 +13,23 @@ function setSolverParams(dfg::LightDFG, solverParams::P) where P <: AbstractPara
   dfg.solverParams = solverParams
 end
 
+# Get user, robot, and session "small" data.
+getUserData(dfg::LightDFG)::Dict{Symbol, String} = return dfg.userData
+function setUserData(dfg::LightDFG, data::Dict{Symbol, String})::Bool
+	dfg.userData = data
+	return true
+end
```
Affie: Should we not `merge!`, I guess then `addUserData!` will be a more appropriate name?
It would spare the garbage collector if new pointers are avoided as much as possible.

GearsAD: Yeah, in this case I feel it's atomic? User is responsible for merging because it's a block of data. This is in case you remove something and then setUserData, a merge would just replace it.

Affie: Agree, set sounds like it should replace:
How would you suggest it is used to add to for example:
```julia
userdata = getUserData(dfg) 
push!(userdata, :key=>"value")
setUserData(dfg, userdata) # guess this step will only be needed for the cloud
```
What about the dictionary api since we are working with one: popUserData!, getUserData, deleteUserData!, pushUserData!? 
Example then becomes:
```julia
pushUserData!(dfg, :key=>"value")
```
@dehann, do you have any preferences.

dehann: I like what Johan suggested here.  Assuming user/robot/session are always string in this case.

PR #148, opened 2019-10-10
address issue #147 (standard PPE)

### LINKED ISSUE #147: Designing PPE `VariableEstimate`
Parametric Point Estimates (PPE) are stored in `VariableEstimate`, but there is a question of what data should be stored with it?  

Keeping in mind the data is stored in dictionary, example from current design:
```julia
 var.estimateDict[:default] = Dict{Symbol, VariableEstimate}(
    :max => VariableEstimate(:default, :max, varMax),
    :mean => VariableEstimate(:default, :mean, varMean),
    :modefit => VariableEstimate(:default, :modefit, varModefit))
```

Suggestions are:
- [x] pppeType, `:mean / :max / :modefit`
- [x] actual value 
- [ ] softtypes which can point to (`:Pose2`) or contain manifolds (`Pose2().manifolds`)
- [x] `solveKey`
- [x] `lastupdated::DateTime`
...

Current here:
https://github.com/JuliaRobotics/DistributedFactorGraphs.jl/blob/bca01e6188ef206122a7c9a51378ca3d0f1e0b8c/src/entities/DFGVariable.jl#L76-L82

dehann:
resolves #147

### DIFF
--- a/src/entities/DFGVariable.jl
+++ b/src/entities/DFGVariable.jl
@@ -117,6 +117,14 @@ timestamp(v::DFGVariable) = v.timestamp
 tags(v::DFGVariable) = v.tags
 estimates(v::DFGVariable) = v.estimateDict
 estimate(v::DFGVariable, key::Symbol=:default) = haskey(v.estimateDict, key) ? v.estimateDict[key] : nothing
+
+"""
+    $SIGNATURES
+
+Retrieve the soft type name symbol for a DFGVariable or DFGVariableSummary. ie :Point2, Pose2, etc.
+"""
+softtype(v::DFGVariable)::Symbol = Symbol(typeof(getSofttype(v)))
+
 """
     $SIGNATURES
 
--- a/src/entities/DFGVariable.jl
+++ b/src/entities/DFGVariable.jl
@@ -73,6 +73,8 @@ mutable struct PackedVariableNodeData
                          x15::Bool ) = new(x1,x2,x3,x4,x5,x6,x7,x8,x9,x10,x11,x12,x13,x14,x15)
 end
 
+
+abstract type AbstractVariableEstimate end
 """
     $TYPEDEF
 
@@ -84,7 +86,7 @@ Notes
 - `estimate` is the actual numerical estimate value,
 - Additional information such as how the data is represented (ie softtype) is stored alongside this data container in the `DFGVariableSummary` container.
 """
-struct VariableEstimate
+struct VariableEstimate <: AbstractVariableEstimate
   solverKey::Symbol
   ppeType::Symbol
   estimate::Vector{Float64}
@@ -100,7 +102,7 @@ mutable struct DFGVariable <: AbstractDFGVariable
     label::Symbol
     timestamp::DateTime
     tags::Vector{Symbol}
-    estimateDict::Dict{Symbol, Dict{Symbol, VariableEstimate}}
+    estimateDict::Dict{Symbol, Dict{Symbol, <: AbstractVariableEstimate}}
     solverDataDict::Dict{Symbol, VariableNodeData}
     smallData::Dict{String, String}
     bigData::Any
@@ -160,7 +162,7 @@ mutable struct DFGVariableSummary <: AbstractDFGVariable
     label::Symbol
     timestamp::DateTime
     tags::Vector{Symbol}
-    estimateDict::Dict{Symbol, Dict{Symbol, VariableEstimate}}
+    estimateDict::Dict{Symbol, Dict{Symbol, <:AbstractVariableEstimate}}
     softtypename::Symbol
     _internalId::Int64
 end
--- a/src/entities/DFGVariable.jl
+++ b/src/entities/DFGVariable.jl
@@ -10,7 +10,7 @@ mutable struct VariableNodeData
   eliminated::Bool
   BayesNetVertID::Symbol #  Union{Nothing, }
   separator::Array{Symbol,1}
-  softtype
+  softtype  # Perhaps this should move up to DFGVariable level
   initialized::Bool
   inferdim::Float64
   ismargin::Bool
@@ -73,9 +73,20 @@ mutable struct PackedVariableNodeData
                          x15::Bool ) = new(x1,x2,x3,x4,x5,x6,x7,x8,x9,x10,x11,x12,x13,x14,x15)
 end
 
+"""
+    $TYPEDEF
+
+Data container to store Parameteric Point Estimate (PPE) from a variety of types.
+
+Notes
+- `ppeType` is something like `:max/:mean/:modefit` etc.
+- `solveKey` is from super-solve concept, starting with `:default`,
+- `estimate` is the actual numerical estimate value,
+- Additional information such as how the data is represented (ie softtype) is stored alongside this data container in the `DFGVariableSummary` container.
+"""
 struct VariableEstimate
   solverKey::Symbol
-  type::Symbol
+  ppeType::Symbol
   estimate::Vector{Float64}
   lastUpdatedTimestamp::DateTime
   VariableEstimate(solverKey::Symbol, type::Symbol, estimate::Vector{Float64}, lastUpdatedTimestamp::DateTime=now()) = new(solverKey, type, estimate, lastUpdatedTimestamp)
@@ -142,6 +153,7 @@ mutable struct DFGVariableSummary <: AbstractDFGVariable
     timestamp::DateTime
     tags::Vector{Symbol}
     estimateDict::Dict{Symbol, Dict{Symbol, VariableEstimate}}
+    softtypename::Symbol
     _internalId::Int64
 end
 label(v::DFGVariableSummary) = v.label
@@ -149,4 +161,5 @@ timestamp(v::DFGVariableSummary) = v.timestamp
 tags(v::DFGVariableSummary) = v.tags
 estimates(v::DFGVariableSummary) = v.estimateDict
 estimate(v::DFGVariableSummary, key::Symbol=:default) = haskey(v.estimateDict, key) ? v.estimateDict[key] : nothing
+softtype(v::DFGVariableSummary)::Symbol = v.softtypename
 internalId(v::DFGVariableSummary) = v._internalId
--- a/src/Common.jl
+++ b/src/Common.jl
@@ -129,7 +129,7 @@ function getfnctype(data::GenericFunctionNodeData)
   return data.fnc.usrfnc!
 end
 function getfnctype(fact::DFGFactor; solveKey::Symbol=:default)
-  data = getData(fact) # TODO , solveKey=solveKey)
+  data = solverData(fact) # TODO , solveKey=solveKey)
   return getfnctype(data)
 end
 function getfnctype(dfg::T, lbl::Symbol; solveKey::Symbol=:default) where T <: AbstractDFG
@@ -145,7 +145,7 @@ Notes
 - Replaces older `getfnctype`.
 """
 getFactorType(data::GenericFunctionNodeData) = data.fnc.usrfnc!
-getFactorType(fct::DFGFactor) = getFactorType(getData(fct))
+getFactorType(fct::DFGFactor) = getFactorType(solverData(fct))
 function getFactorType(dfg::G, lbl::Symbol) where G <: AbstractDFG
   getFactorType(getFactor(dfg, lbl))
 end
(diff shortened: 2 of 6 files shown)


src/entities/DFGVariable.jl
```diff
@@ -10,7 +10,7 @@ mutable struct VariableNodeData
   eliminated::Bool
   BayesNetVertID::Symbol #  Union{Nothing, }
   separator::Array{Symbol,1}
-  softtype
+  softtype  # Perhaps this should move up to DFGVariable level
```
Affie: I would think only the `softtypename::Symbol`
but I added a softtype(v::DFGVariable) to maintain compatibility.

src/entities/DFGVariable.jl
```diff
@@ -73,9 +73,20 @@ mutable struct PackedVariableNodeData
...
+Notes
+- `ppeType` is something like `:max/:mean/:modefit` etc.
+- `solveKey` is from super-solve concept, starting with `:default`,
+- `estimate` is the actual numerical estimate value,
+- Additional information such as how the data is represented (ie softtype) is stored alongside this data container in the `DFGVariableSummary` container.
+"""
 struct VariableEstimate
   solverKey::Symbol
-  type::Symbol
+  ppeType::Symbol
```
Affie: I'm happy with the name change, less confusing.

Affie (approved):
See my changes for tests also

src/entities/DFGVariable.jl
dehann: see #149

dehann:
I'm really worried about performance hit of using `<:AbstractVariableEstimate` this way.  DFGVariable is used a lot.  Also not sure we want to template all of `DFGVariable{T <: AbstractVariableEstimate}` -- seems like a lot of excess in other places for something fairly small (PPE)?

src/entities/DFGVariable.jl
dehann: > should DFG not export an abstract type with IIF InferenceType inheriting from it?

DFG already exports `InferenceType` and `InferenceVariable`.  `InferenceType` is already an abstract type, so dont think we need to add another layer. https://github.com/JuliaRobotics/DistributedFactorGraphs.jl/issues/149#issuecomment-540642450

Affie:
> I'm really worried about performance hit of using `<:AbstractVariableEstimate` this way.  DFGVariable is used a lot.  Also not sure we want to template all of `DFGVariable{T <: AbstractVariableEstimate}` -- seems like a lot of excess in other places for something fairly small (PPE)?

Sorry, I don’t understand. I didn’t make it a container of abstract variables because you were worried about performance. And didn't make parametric type, because i agree its unnecessary complications and restricting. Guess I need to learn some more julia then.

dehann:
I think we are close to a decision (Sams suggestions seems like best mix of both).  Dict of VariableEstimate containing a Dict).

> I didn’t make it a container of abstract variables because you were worried about performance.

We can continue discussion post decision to make sure we are speaking the same language and that you are happy with the result -- when you did this in af107a6

> `estimateDict::Dict{Symbol, Dict{Symbol, <: AbstractVariableEstimate}}`

isn't that `a container of abstract variables`??  It like an abstract container to me...

Affie:
I don't mind the new changes sam is proposing. Just note they will be breaking, but rather sooner than later. 
Just for my understanding of optimal code.
It is not abstract in this case:
```julia
abstract type AbstractTypeAB end

struct TypeA <: AbstractTypeAB
    a::Symbol
    b::Int
end

struct TypeB <: AbstractTypeAB
    a::Symbol
    b::Int
end

struct Option1 
    s::Symbol
    d::Dict{Symbol, AbstractTypeAB}
end

struct Option2 
    s::Symbol
    d::Dict{Symbol, <:AbstractTypeAB}
end

julia> o1 = Option1(:bla, Dict{Symbol,TypeA}())
Option1(:bla, Dict{Symbol,AbstractTypeAB}())

julia> push!(o1.d, :a=>TypeA(:a,1))
Dict{Symbol,AbstractTypeAB} with 1 entry:
  :a => TypeA(:a, 1)

julia> push!(o1.d, :b=>TypeB(:b,1))
Dict{Symbol,AbstractTypeAB} with 2 entries:
  :a => TypeA(:a, 1)
  :b => TypeB(:b, 1)

julia> o2 = Option2(:bla, Dict{Symbol,TypeA}())
Option2(:bla, Dict{Symbol,TypeA}())

julia> push!(o2.d, :a=>TypeA(:a,1))
Dict{Symbol,TypeA} with 1 entry:
  :a => TypeA(:a, 1)

julia> push!(o2.d, :b=>TypeB(:b,1))
ERROR: MethodError: Cannot `convert` an object of type TypeB to an object of type TypeA
Closest candidates are:
  convert(::Type{T}, ::T) where T at essentials.jl:167
  TypeA(::Any, ::Any) at untitled-b44b38864b48aa54882edecfec5246ba:5
Stacktrace:
 [1] setindex!(::Dict{Symbol,TypeA}, ::TypeB, ::Symbol) at ./dict.jl:380
 [2] push!(::Dict{Symbol,TypeA}, ::Pair{Symbol,TypeB}) at ./abstractdict.jl:479
 [3] top-level scope at none:0
```
But I guess you can force it with the second (outer) dict:
```julia
:default=>Dict{Symbol, VariableEstimate}} # can only contain VariableEstimate
:other=>Dict{Symbol, OtherVariableEstimateType}}#can only contain OtherVariableEstimateType
but Dict{Symbol, Dict{Symbol, <: AbstractVariableEstimate}}  would then contain both. 
```
I still don't see the problem with it. I also think it happens in the Variables dictionary, since not all variables are of the same type. ie pose point

dehann:
My opinion is that this is the part that is bad for performance:
```julia
Dict{Symbol, <: AbstractVariableEstimate}}
```

dehann:
Functionally yes they will work, but that requires dynamic type inference.

Affie:
Why is it different than templating? It also fixes the type when it is created, after the type cannot be changed and can only be of TypeA in this example.

dehann:
> Why is it different than templating?

Julia syntax specific thing.  

### `Option1` case
How does the compiler know in this case
```julia
struct Option1 
    s::Symbol
    d::Dict{Symbol, AbstractTypeAB}
end
```
when you are going to do this:
```julia
julia> push!(o1.d, :b=>TypeB(:b,1))
Dict{Symbol,AbstractTypeAB} with 2 entries:
  :a => TypeA(:a, 1)
  :b => TypeB(:b, 1)
```
hence dynamic type inference will always be needed, even if you just use it with `TypeA`.  The very next one might be `TypeB`.

### `Option2` case, 
There is some wiggle room:
```julia
julia> o2 = Option2(:bla, Dict{Symbol,TypeA}())
Option2(:bla, Dict{Symbol,TypeA}())

julia> push!(o2.d, :a=>TypeA(:a,1))
Dict{Symbol,TypeA} with 1 entry:
  :a => TypeA(:a, 1)

julia> push!(o2.d, :b=>TypeB(:b,1))
ERROR: MethodError: Cannot `convert` an object of type TypeB to an object of type TypeA
Closest candidates are:
  convert(::Type{T}, ::T) where T at essentials.jl:167
  TypeA(::Any, ::Any) at untitled-b44b38864b48aa54882edecfec5246ba:5
```

I'm not sure if the compiler will interpret that as an anonymous static parameter or the same as `Option1`.  They Julia guys will know.  Perhaps I could ask for a higher opinion from @andreasnoack (Hi!) if he knows about Option1 vs Option2 in this example: https://github.com/JuliaRobotics/DistributedFactorGraphs.jl/pull/148#issuecomment-541536774

Historically, I have always understood static type parameters (templating) as a requirement to avoid dynamic type inference.  The Option2 example could well be up to how successful static type inference is at resolving static parameters, and maybe newer Julia's are better than previous versions.

Affie:
Thanks, I think you finally get why I struggle to understand you. 
I have it as an anonymous parameter, type generated is: `d::Dict{Symbol,#s142} where #s142<:AbstractTypeAB`

Affie:
Maybe this is related: 
https://docs.julialang.org/en/v1/manual/style-guide/#Don't-use-unnecessary-static-parameters-1
> There is no performance difference

dehann:
if anonymous parameter are reliable in all cases (not just small ones) --- e.g. `#s142<:AbstractTypeAB` --- then great i learned something, thanks!  I do recall specifically this being a problem a couple months back, and i just haven't kept up with developments.

dehann:
> Maybe this is related:

Sure, but as before -- that is for `function` (aka member) definitions, not `struct` definition.  Two very different beasts.  Yes, will be good if the language syntax is consistent.  My push for explicit static parameter on `struct` is conservative for performance reasons.

dehann (commented):
I was not aware of anonymous static types in Julia, so Johan has convinced me of the alternative.

Affie:
Closed by design decision from #147 [last few comments](https://github.com/JuliaRobotics/DistributedFactorGraphs.jl/issues/147#issuecomment-542091523)

andreasnoack:
Sorry for the slow reply here. Is the main worry the overhead from dynamic dispatch because the compiler can't infer the types or is the worry excessive compile time because of the use of type parameters? Let me try to modify the example slightly to highlight what's inferred and what is not. Here are three Option structs that all store dicts with `Real` values
```julia
julia> struct Option1
         x::Dict{Symbol,Real}
       end

julia> struct Option2
         x::Dict{Symbol,<:Real}
       end

julia> struct Option3{T}
         x::Dict{Symbol,T}
       end
```
To see what gets inferred and what doesn't, consider the two helper functions
```julia
julia> foo1(o) = o.x
foo1 (generic function with 1 method)

julia> foo2(d) = d[:a]
foo2 (generic function with 1 method)

julia> foo3(o) = o.x[:a]
foo3 (generic function with 1 method)
```
and the dict
```julia
julia> d = Dict(:a =>1)
Dict{Symbol,Int64} with 1 entry:
  :a => 1
```
From
```julia
julia> @code_warntype foo1(Option1(d))
Variables
  #self#::Core.Compiler.Const(foo1, false)
  o::Option1

Body::Dict{Symbol,Real}
1 ─ %1 = Base.getproperty(o, :x)::Dict{Symbol,Real}
└──      return %1

julia> @code_warntype foo1(Option2(d))
Variables
  #self#::Core.Compiler.Const(foo1, false)
  o::Option2

Body::Dict{Symbol,#s1} where #s1<:Real
1 ─ %1 = Base.getproperty(o, :x)::Dict{Symbol,#s1} where #s1<:Real
└──      return %1

julia> @code_warntype foo1(Option3(d))
Variables
  #self#::Core.Compiler.Const(foo1, false)
  o::Option3{Int64}

Body::Dict{Symbol,Int64}
1 ─ %1 = Base.getproperty(o, :x)::Dict{Symbol,Int64}
└──      return %1
```
we can see that 1 and 3 are inferred when calling `foo1`. However, from
```julia
julia> @code_warntype foo2(Option2(d).x)
Variables
  #self#::Core.Compiler.Const(foo2, false)
  d::Dict{Symbol,Int64}

Body::Int64
1 ─ %1 = Base.getindex(d, :a)::Int64
└──      return %1

julia> @code_warntype foo2(Option3(d).x)
Variables
  #self#::Core.Compiler.Const(foo2, false)
  d::Dict{Symbol,Int64}

Body::Int64
1 ─ %1 = Base.getindex(d, :a)::Int64
└──      return %1
```
we see that extracting the element from the dict is only inferred in 2 and 3. Finally, since failure of inference propagates we get
```julia
julia> @code_warntype foo3(Option1(d))
Variables
  #self#::Core.Compiler.Const(foo3, false)
  o::Option1

Body::Real
1 ─ %1 = Base.getproperty(o, :x)::Dict{Symbol,Real}
│   %2 = Base.getindex(%1, :a)::Real
└──      return %2

julia> @code_warntype foo3(Option2(d))
Variables
  #self#::Core.Compiler.Const(foo3, false)
  o::Option2

Body::Real
1 ─ %1 = Base.getproperty(o, :x)::Dict{Symbol,#s1} where #s1<:Real
│   %2 = Base.getindex(%1, :a)::Real
└──      return %2

julia> @code_warntype foo3(Option3(d))
Variables
  #self#::Core.Compiler.Const(foo3, false)
  o::Option3{Int64}

Body::Int64
1 ─ %1 = Base.getproperty(o, :x)::Dict{Symbol,Int64}
│   %2 = Base.getindex(%1, :a)::Int64
└──      return %2
```
i.e. only 3 is inferred.

If you want to avoid the dynamic dispatch, you should therefore use 3. However, it might not always be beneficial since it can cause extra compile time. In particular if the types can't be inferred. If the is a concern then it might actually be worth considering
```julia
julia> struct Option4
         x::Dict{Symbol,Any}
       end
```
since it might have the minimal compile overhead.

GearsAD:
@andreasnoack Thanks for the great writeup!

PR #158, opened 2019-10-14
Example using a blend of hard type metadata and dicts for adding new …

GearsAD:
…estimates

### DIFF
--- a/src/entities/DFGVariable.jl
+++ b/src/entities/DFGVariable.jl
@@ -78,24 +78,16 @@ mutable struct PackedVariableNodeData
 end
 
 
-abstract type AbstractVariableEstimate end
 """
     $TYPEDEF
 
 Data container to store Parameteric Point Estimate (PPE) from a variety of types.
-
-Notes
-- `ppeType` is something like `:max/:mean/:modefit` etc.
-- `solveKey` is from super-solve concept, starting with `:default`,
-- `estimate` is the actual numerical estimate value,
-- Additional information such as how the data is represented (ie softtype) is stored alongside this data container in the `DFGVariableSummary` container.
 """
-struct VariableEstimate <: AbstractVariableEstimate
-  solverKey::Symbol
-  ppeType::Symbol
-  estimate::Vector{Float64}
+struct PointParametricEstimates
+  # Put other fields here.
+  estimates::Dict{Symbol, Vector{Float64}}
   lastUpdatedTimestamp::DateTime
-  VariableEstimate(solverKey::Symbol, type::Symbol, estimate::Vector{Float64}, lastUpdatedTimestamp::DateTime=now()) = new(solverKey, type, estimate, lastUpdatedTimestamp)
+  PointParametricEstimates(lastUpdatedTimestamp::DateTime=now()) = new(Dict{Symbol, Vector{Float64}}(), lastUpdatedTimestamp)
 end
 
 """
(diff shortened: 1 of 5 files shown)


dehann (approved):

dehann:
Looks good to me!

dehann:
part of #147 discussion

dehann:
want to make sure @Affie is happy before we merge the twig, and then to master

Affie:
You can go ahead, I'm not an expert when it comes to julia.
The only thing I've found with types represented as dictionaries is its harder to use. 
Also please define the fields somehow (maybe in documentation), so I know for example :max will be there and the name won't change to something else. eg. :maximum

dehann:
unlikely to have a name change from (eg :max to :maximum), but if it does occur we will do so with deprecation cycle and minor version numbers.

Affie (commented):
Maybe just hold off on it not to break IIF and first do the branch. and this for v0.5.0

dehann:
TLDR; hold off on merging this until we have spoken again please.

hi @GearsAD , Johan convinced me on using `Dict{Symbol, <: AnAbstractType}` given new and improved type inference with anonymous static parameters. https://github.com/JuliaRobotics/DistributedFactorGraphs.jl/pull/148#issuecomment-541536774

So shall we change that now or later?  I told Johan later, but I thought this was already merged.  Perhaps we can just do it now before merging this.  #148 already has it the way Johan suggested, we can maybe just introduce the name change to `PointParametericEstimate` etc.

src/entities/DFGVariable.jl
```diff
@@ -106,15 +98,15 @@ mutable struct DFGVariable <: AbstractDFGVariable
     label::Symbol
     timestamp::DateTime
     tags::Vector{Symbol}
-    estimateDict::Dict{Symbol, Dict{Symbol, <: AbstractVariableEstimate}}
+    estimateDict::Dict{Symbol, PointParametricEstimates}
```
dehann: `estimateDict::Dict{Symbol, <: AbstractPointParametric}`

or whatever the chosen name for `AbstractPointParametric` is in the end

src/entities/DFGVariable.jl
```diff
@@ -78,24 +78,16 @@ mutable struct PackedVariableNodeData
...
-- `solveKey` is from super-solve concept, starting with `:default`,
-- `estimate` is the actual numerical estimate value,
-- Additional information such as how the data is represented (ie softtype) is stored alongside this data container in the `DFGVariableSummary` container.
-"""
-struct VariableEstimate <: AbstractVariableEstimate
-  solverKey::Symbol
-  ppeType::Symbol
-  estimate::Vector{Float64}
+"""
+struct PointParametricEstimates
```
dehann: `struct PointParametricEstimates <: AbstractPointParametric` as originally suggested by @Affie.

or whatever the chosen name for `AbstractPointParametric` is in the end

dehann (changes requested):

src/entities/DFGVariable.jl
dehann: from #148 

`<: AbstractVariableEstimate`

src/entities/DFGVariable.jl
dehann: from #148

`<: AbstractVariableEstimate`

Affie:
Closed by design decision from #147 [last few comments](https://github.com/JuliaRobotics/DistributedFactorGraphs.jl/issues/147#issuecomment-542091523)

PR #162, opened 2019-10-15
Feature/skeleton dfg

Affie:
# Notes
- Adds the skeleton types, that are just labels and tags of variables and factors.


dehann:
Are we targeting this PR for v0.5.0, or later in v0.5.x?

Affie:
We can do v0.5.0, I'm just not completely convinced about a few things. 
1. I would like the summary and skeleton types to be immutable from the start.
    - So then the difference in `addVariable` and `addFactor`  bugs the OCD part in me.
    - First choice for me is to set everything at construction. Otherwise what is safest and most efficient way of  modifying the immutable fields. eg a `Vector{Symbol}`, resize or whole new struct, kind of thing.     
2. Does it need tags, if I look at a graph db it looks like a yes. It also improves comparability with abstraction layer.

Affie:
I think tags are in then. 
I think .= should still be faster than mutable. 
Most types in julia are mutable, so to have the summary immutable should not be that hard

test/LightDFGSummaryTypes.jl
```diff
@@ -112,58 +117,60 @@ end
...
-        varest1.solverKey == varest2.solverKey || return false
-        varest1.estimate == varest2.estimate || return false
-        return true
+    if VARTYPE == DFGVariableSummary
+        global dfg
+        #get the variable
+        var = getVariable(dfg, :a)
+        #make a copy and simulate external changes
+        newvar = deepcopy(var)
+        estimates(newvar)[:default] = Dict{Symbol, VariableEstimate}(
```
dehann: would probably have to change to `AbstractPointParametricEst` ?

test/LightDFGSummaryTypes.jl
```diff
@@ -112,58 +117,60 @@ end
...
+
+        # Delete :default and replace to see if new ones can be added
+        delete!(estimates(newvar), :default)
+        estimates(newvar)[:second] = Dict{Symbol, VariableEstimate}(
+            :max => VariableEstimate(:default, :max, [10.0]),
+            :mean => VariableEstimate(:default, :mean, [5.0]),
+            :ppe => VariableEstimate(:default, :ppe, [7.0]))
+
+        # Persist to the original variable.
+        mergeUpdateVariableSolverData!(dfg, newvar)
```
dehann: just check if this has a relation to discussion at #182 ?

dehann (changes requested):
principle looks good, I like it!  Changes requested are meant positive.

test/LightDFGSummaryTypes.jl
Affie: Jip, but both in branches. I think one of them needs to be merged first, then it can be updated. To get travis to pass

test/LightDFGSummaryTypes.jl
Affie: Jip.  We had a different view on #166 how it worked. I used merge in LightDFG and GraphDFG, and sam replaced. So he changed it to mergeupdate. I think we can sort it out in #182

test/LightDFGSummaryTypes.jl
dehann: suggestion: maybe merge the other PR to master first, then pull master in here and get tests to pass here and finally merge this PR after?

dehann: im guessing #183

Affie: Jip

GearsAD: Okay so first 183? We can merge that this weekend :+1:

GearsAD (approved):

test/LightDFGSummaryTypes.jl
GearsAD: Merged and updated, let me know if I missed anything.

Affie: LGTM

dehann (approved):

src/entities/DFGFactor.jl
```diff
@@ -103,3 +103,19 @@ label(f::DFGFactorSummary) = f.label
...
+
+
+# SKELETON DFG
+"""
+	$(TYPEDEF)
+Skeleton variable with essentials.
+"""
+struct SkeletonDFGVariable <: AbstractDFGVariable
+	label::Symbol
+	tags::Vector{Symbol}
```
dehann: skeleton does not have variable order, but i think that should be okay.

Affie: It has `_variableOrderSymbols` in DFGFactor

PR #132, opened 2019-10-02
Feature/iif tests

GearsAD:
# Notes
- Adds IncrementalInference test job
- Testing DFG with Julia > 1.2 will run those tests
- Neo4J set up for Travis
- Start of CGDFG cleanup

PR #167, opened 2019-10-18
megamerge - trying to tie up loose ends

Affie:
Merges:
- #132 and #165 iif tests
- #148 renames and inheritance, does not cause a breaking change in IIF.
- #140 + additional (still outstanding on cloud, but covered in tests).
- #136, just to have consistancy until we change it in a breaking change to #134 
- #131 big data draft (still outstanding on cloud).

Plus addition bug fixes and tests.

To get everyone on the same page, then we can start from here.


src/DistributedFactorGraphs.jl
```diff
@@ -67,18 +72,26 @@ include("LightDFG/LightDFG.jl")
...
         include("DFGPlots/DFGPlots.jl")
         @reexport using .DFGPlots
     end
 
 end
 
+#FIXME still can't figure @require out
+# Include the Cloudgraphs API
+include("CloudGraphsDFG/CloudGraphsDFG.jl")
+
```
Affie: This is just a temporary measure until someone can tell me how to get it to work.

src/entities/DFGVariable.jl
```diff
@@ -77,39 +63,69 @@ mutable struct PackedVariableNodeData
...
 mutable struct DFGVariable <: AbstractDFGVariable
     label::Symbol
     timestamp::DateTime
     tags::Vector{Symbol}
-    estimateDict::Dict{Symbol, Dict{Symbol, VariableEstimate}}
+    estimateDict::Dict{Symbol, Dict{Symbol, <: AbstractVariableEstimate}}
     solverDataDict::Dict{Symbol, VariableNodeData}
     smallData::Dict{String, String}
-    bigData::Any
+    bigData::Dict{Symbol, AbstractBigDataEntry}
```
Affie: better than Any

src/entities/DFGVariable.jl
```diff
@@ -122,7 +138,12 @@ solverData(v::DFGVariable, key::Symbol=:default) = haskey(v.solverDataDict, key)
 Retrieve data structure stored in a variable.
 """
 function getData(v::DFGVariable; solveKey::Symbol=:default)::VariableNodeData
-  @warn "getData is deprecated, please use solverData()"
+  #FIXME but back in later, it just slows everything down
+  if !(@isdefined getDataWarnOnce)
+    @warn "getData is deprecated, please use solverData(), future warnings in getData is suppressed"
+    global getDataWarnOnce = true
+  end
+  # @warn "getData is deprecated, please use solverData()"
```
Affie: Got tired of messages, we can remove this since IIF was updated.

src/services/AbstractDFG.jl
```diff
@@ -49,6 +49,50 @@ function setSolverParams(dfg::G, solverParams::T) where {G <: AbstractDFG, T <:
...
+end
+
+pushUserData!(dfg::AbstractDFG, pair::Pair{Symbol,String}) = push!(dfg.userData, pair)
+pushRobotData!(dfg::AbstractDFG, pair::Pair{Symbol,String}) = push!(dfg.userData, pair)
+pushSessionData!(dfg::AbstractDFG, pair::Pair{Symbol,String}) = push!(dfg.userData, pair)
+
+popUserData!(dfg::AbstractDFG, key::Symbol) = pop!(dfg.userData, key)
+popRobotData!(dfg::AbstractDFG, key::Symbol) = pop!(dfg.userData, key)
+popSessionData!(dfg::AbstractDFG, key::Symbol) = pop!(dfg.userData, key)
+
```
Affie: This should still be overwritten by CloudDFG, maybe FileDFG also

src/services/AbstractDFG.jl
```diff
@@ -449,7 +493,7 @@ Return boolean whether a factor `label` is present in `<:AbstractDFG`.
 """
 function hasFactor(dfg::G, label::Symbol)::Bool where {G <: AbstractDFG}
 	@warn "hasFactor() deprecated, please use exists()"
-	return haskey(dfg.labelDict, label)
+	return exists(dfg, label)
```
Affie: I don't use labeldict in LighDFG, seems like its replaced with exist everywhere.

src/services/AbstractDFG.jl
```diff
@@ -472,14 +516,16 @@ Notes:
 - used by both factor graph variable and Bayes tree clique logic.
 """
 function isInitialized(var::DFGVariable; key::Symbol=:default)::Bool
-  solverData(var, key) != nothing && return solverData(var, key).initialized
-  return false
-end
-function isInitialized(fct::DFGFactor; key::Symbol=:default)::Bool
-  solverData(var, key) != nothing && return solverData(fct, key).initialized
-  return false
```
Affie: DFG factor does not have this field if I remember correctly

src/services/AbstractDFG.jl
```diff
@@ -472,14 +516,16 @@ Notes:
...
+  	data = solverData(var, key)
+  	if data == nothing
+		@error "Variable does not have solver data $(key)"
```
Affie: I thought an @error should be raised, because something is seriously wrong and a false return just lets me think its not initialized.

test/IIF4DoorsTests.jl
```diff
@@ -0,0 +1,81 @@
+using GraphPlot
+using Neo4j
+using DistributedFactorGraphs
+using IncrementalInference
+
+@testset "test fourdoor early example" begin
```
Affie: Left this as it can be used as template for end to end tests.

test/runtests.jl
```diff
@@ -33,3 +41,37 @@ end
     @info "Testing LightDFG Variable and Factor Subtypes"
     include("LightDFGSummaryTypes.jl")
 end
+
+
+if get(ENV, "IIF_TEST", "") == "true"
```
Affie: This is important to keep

test/iifInterfaceTests.jl
```diff
@@ -0,0 +1,437 @@
...
+end
+
+
+# Now make a complex graph for connectivity tests
+numNodes = 10
+#the deletions in last test should have cleared out the fg
+# dfg = DistributedFactorGraphs._getDuplicatedEmptyDFG(dfg)
+# if typeof(dfg) <: CloudGraphsDFG
+#     clearSession!!(dfg)
+# end
```
Affie: I serves as additional tests not to clear the session

test/iifInterfaceTests.jl
```diff
@@ -0,0 +1,437 @@
...
+                                            nothing,
+                                            IncrementalInference.decodePackedType,
+                                            IncrementalInference.rebuildFactorMetadata!,
+                                            solverParams=SolverParams())
+    else
+        dfg2 = T()
+    end
+
+    # Build a new in-memory IIF graph to transfer into the new graph.
+    iiffg = initfg()
```
Affie: Please leave this in-tact. I want to continue extending tests to work between different graphs.

Affie: we should rather create a new test for the _get duplicated graph.

src/entities/DFGVariable.jl
dehann: funny story, the remaining message now all come from within DFG

src/services/AbstractDFG.jl
dehann: oh, good ot know thanks.  I have been using `haskey`

src/services/AbstractDFG.jl
dehann: DFGFactor has thus far not required an `initialized` field, but does have an `eliminated` field which operates roughly in the same area in IIF.

src/services/AbstractDFG.jl
dehann: just making sure you are confortable with difference between `error(..)` and `@error`?

test/runtests.jl
dehann: okay, think @GearsAD was using internal Julia function to test if IIF is defined and then conditions on that.  Either way is good at this point.

src/entities/DFGVariable.jl
Affie: It should be fixed now

test/iifInterfaceTests.jl
dehann: good point, I frequency use `deepcopy(fg)` all over the place.  A way to clone or duplicate factor graphs will be pretty important i think

.travis.yml
```diff
@@ -13,37 +18,35 @@ julia:
...
 jobs:
   include:
+    - julia: 1.2
+      env: IIF_TEST=true
     - stage: "Documentation"
       julia: 1.0
       os: linux
       script:
-        - julia -e 'import Pkg; Pkg.add("Documenter"); Pkg.add("DataFrames")'
+        - julia -e 'import Pkg; Pkg.add("Documenter"); Pkg.add("Neo4j"); Pkg.add("GraphPlot")'
```
dehann: there is a way to do this with the TOML files perhaps, but not serious.

src/services/AbstractDFG.jl
Affie: Should we open an issue for this? do you need something similar for eliminated, `isEliminated`. I thought it was a copy and pate error. since the field does not exists.

src/services/AbstractDFG.jl
Affie: haha, I do. We should not be using error() in any case, it throws an generic error.

Affie: Didn't want to mess with the existing behavior as it is a breaking change.

test/runtests.jl
Affie: This allows for easier matrix setup in travis.

Affie: just din't want a merge conflict to revert it back.

.travis.yml
Affie: Don't know how, I only know of  test dependencies.

test/runtests.jl
GearsAD: Yeah the Pkg check works a little cleaner than an environment variable but merging as is for now.

Affie: Maybe i should rephrase. It’s important to keep the way i changed the travis setup. It fixed 4 things, so as long as they stay the same i don’t mind if iif is looked for. Fixed:
- iif test was done on latest tag
- jobs are now run in parallel 
- failures allowed to not break on circular dependancies 
-coverage report uploaded on pass

src/entities/DFGVariable.jl
dehann: thanks for the fix!

src/services/AbstractDFG.jl
dehann: I'm in two minds about an issue here.  Yes it was a copy paste, but maybe when we get to the performance improvements the concept of and "initialized" factor starts to exist, things like allocate the predicted memory for in-place operations, etc.  Do you have any strong feelings either way?

src/services/AbstractDFG.jl
dehann: cool, so to be sure -- how do you capture `@error` calls?

test/IIF4DoorsTests.jl
dehann: separate discussion on how cyclic IIF dependency can be included for tests -- decision thus far was to add conditional tests.

dehann:
Thanks again, this is great!

src/services/AbstractDFG.jl
Affie: You don’t, user reads in REPL or log.

test/runtests.jl
dehann: @Affie, i'm all good with what you listed here.  So dont have to fiddle with that.

.travis.yml
dehann: will figure out some other time.

src/entities/DFGVariable.jl
dehann: agree!

.travis.yml
Affie: I fixed it in #180, sorry to abuse your pr.

Issue #166, opened 2019-10-18
updateVariableSolverData! behavior


Affie:
@GearsAD, From tests I see we may have different views on what  `updateVariableSolverData!` does.
I see update as only add to and not delete existing keys:
- If `var1` has only :default and you update it with `var2` that only has :second. `var1` should get a :second and :default should remain.
- If `var1` has only :default and you update it with `var2` that has :defualt. `var1` :default should be replaced/updated 

https://github.com/JuliaRobotics/DistributedFactorGraphs.jl/blob/b110db99e1a25fedb68f8589d6122cc53ea3aaa7/test/iifInterfaceTests.jl#L200-L203

cc @dehann

GearsAD:
No, because in that case there is no way to delete. We've discussed this and an update is an atomic transaction that performs a complete update, shouldn't make it do unexpected things.

GearsAD:
I'm not sure why you want to make it smarter than a simple CRUD update, I can't see the benefit.

Affie:
I though the idea was to have different solves. So if one dfg runs solve1 and another solve2 what do you keep, solve1 or solve2? I thought both. If you just replace the whole dictionary it will update just one. So it is still a complete update but of one solve and not all solves. 
But let me know if i misunderstand the whole solves thing.

Affie:
> No, because in that case there is no way to delete

Then add a delete. Update should not delete, it should update.

GearsAD:
Okay - so the concern is that on datastructures CRUD behavior is standard, and we're deviating from how you generally treat a dictionary. It's fair that you want to merge, so how about this going to change the name and keep the behavior the same.

Issue #139, opened 2019-10-07
Deprecate `getInnerGraph()`


GearsAD:
Shouldn't be used and available, doesn't serve any purpose.

Issue #130, opened 2019-10-01
getSubgraphAroundNode error in CloudGraphsDFG


Affie:
```
MethodError: no method matching CloudGraphsDFG{SolverParams}(::Connection, ::String, ::String, ::String, ::Nothing, ::Nothing, ::typeof(decodePackedType); solverParams=SolverParams(0, nothing, nothing, false, 99999999999, false, false, true, true, true, false, false, false, false, 500, 100, true, "/tmp/caesar/2019-10-01T15:10:15.226", Dict{Symbol,String}()), description="(Copy of) CloudGraphs DFG", useCache=false)
  Closest candidates are:
    CloudGraphsDFG{SolverParams}(::Connection, ::String, ::String, ::String, ::Any, ::Any, ::Any, !Matched::Any; description, solverParams, useCache) where T<:AbstractParams at /home/johan/.julia/dev/DistributedFactorGraphs/src/CloudGraphsDFG/services/CloudGraphsDFG.jl:31
    CloudGraphsDFG{SolverParams}(::Any, ::Any, ::Any, ::Any, ::Any, ::Any, ::Any, !Matched::Any, !Matched::Any, !Matched::Any, !Matched::Any, !Matched::Any, !Matched::Any, !Matched::Any, !Matched::Any) where T<:AbstractParams at /home/johan/.julia/dev/DistributedFactorGraphs/src/CloudGraphsDFG/entities/CloudGraphsDFG.jl:9 got unsupported keyword arguments "solverParams", "description", "useCache"
    CloudGraphsDFG{SolverParams}(!Matched::String, !Matched::Int64, ::String, ::String, !Matched::String, !Matched::String, !Matched::String, !Matched::Any, !Matched::Any, !Matched::Any, !Matched::Any; description, solverParams, useCache) where T<:AbstractParams at /home/johan/.julia/dev/DistributedFactorGraphs/src/CloudGraphsDFG/services/CloudGraphsDFG.jl:40
  Stacktrace:
   [1] _getDuplicatedEmptyDFG(::CloudGraphsDFG{SolverParams}) at /home/johan/.julia/dev/DistributedFactorGraphs/src/CloudGraphsDFG/services/CloudGraphsDFG.jl:57
[...]

Issue #129, opened 2019-10-01
variable order not preserved in CloudGraphsDFG


Affie:

GearsAD:
Fixed in #170

PR #170, opened 2019-10-19
Fixing CGDFG and 0.5 cleanup

GearsAD:
# Changes
- Fixing CGDFG
  - #128
  - #129 
  - [To be confirmed] #130
- #166 Changed solverUpdate call to be mergeUpdateVariableSolverData! 
- IIF's compare for DFGVariable and DFGFactor have been moved to DFG
- InMemoryDFGTypes now in DFG
- Neo4j is officially a dependency of DFG
- #139 Removing getInnerGraph() 

Everything should work except for data calls (userData, robotData, sessionData), and bigData. 
- They've been excluded from the tests for the moment.
 
# Notes

JuliaRobotics/IncrementalInference.jl#417 must be merged for the tests to pass.


test/Project.toml
```diff
@@ -0,0 +1,18 @@
+[deps]
+Colors = "5ae59095-9a9b-59fe-a467-6f913c188581"
+Dates = "ade2ca70-3891-5945-98fb-dc099432e06a"
+Distributions = "31c24e10-a181-5473-b8eb-7969acd0382f"
+DocStringExtensions = "ffbed154-4ef7-542d-bbb7-c09d3a79fcae"
+GraphPlot = "a2cc645c-3eea-5389-862e-a155d0052231"
+Graphs = "86223c79-3864-5bf0-83f7-82e725a168b6"
+IncrementalInference = "904591bb-b899-562f-9e6f-b8df64c7d480"
```
Affie: IIF as dependency is not needed and can be removed to avoid the whole circular dependency thing.
This test/Project.toml does not have any additional benefit and the normal test section will be supported throughout the julia v1 life cycle.

src/CloudGraphsDFG/services/CommonFunctions.jl
```diff
@@ -101,10 +117,9 @@ end
 $(SIGNATURES)
 Utility function to get a Neo4j node sessions for a robot.
 """
-function _getSessionNeoNodesForRobot(userId::String, robotId::String)::Vector{Neo4j.Node}
+function _getSessionNeoNodesForRobot(neo4jInstance::Neo4jInstance, userId::String, robotId::String)::Vector{Neo4j.Node}
```
Affie: Thanks, I needed this, will give it a try.

test/Project.toml
GearsAD: Yeah, with the environment variable it's not needed.

GearsAD:
Done, sounds like a plan. Also set it to use the IIF branch with DFG 0.5 compatibility, we can remove that once IIF has upgraded.

Project.toml
```diff
@@ -13,6 +13,7 @@ JSON2 = "2535ab7d-5cd8-5a07-80ac-9b1792aadce3"
 LightGraphs = "093fc24a-ae57-5d10-9952-331d41423f4d"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 MetaGraphs = "626554b9-1ddb-594c-aa3c-2596fe9399a5"
+Neo4j = "d2adbeaf-5838-5367-8a2f-e46d570981db"
```
dehann: May want to make this a conditional dependency in the future, if we want to retain RoME on RasPi constraint.  It's definitely okay for now, but probably time for drawings and pizza to figure out where we want the hard and conditional boundaries to be.

PR #173, opened 2019-10-22
Backport of the CGDFG code and unit tests

GearsAD:
# Notes
- Backport of the CloudGraphsDFG code and iifInterfaceTests.jl
- Should still function against IIF 0.7.9

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,6 @@
 language: julia
+sudo: required
+dist: trusty
 
 os:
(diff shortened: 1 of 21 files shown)


Affie:
It looks like I'm somewhere in between with versions. I have most of the functionality of master. But none of the changes that breaks without IIF branch. 
eg I already did the `type -> ppeType` change. It doesn't break iif.

PR #176, opened 2019-10-22
add upper bounds to Project.toml

dehann:
will be needed by zeptodoctor for auto-merging.  Also note new compat syntax coming:

JuliaLang/Pkg.jl#1454


GearsAD:
Great, thanks!

PR #179, opened 2019-10-22
Fixed versions for release

GearsAD:


dehann (approved):
maybe JSON2 also have 0.2 and 0.4 if thats usable, but not serious

PR #180, opened 2019-10-23
allow fast finish on travis testing EDIT: also travis update and documenter cleanup

dehann:


Affie:
I don't know if its in the correct place now. the build is not failing so can't test.
I'm not exactly sure we want this. I would rather have allow failures only on the release branch and IIF test always done on PRs.

dehann:
oh thanks, didnt think of overwriting!

dehann:
> I would rather have allow failures only on the release branch and IIF test always done on PRs.

we can do that I think.  Shall we just update the allow failures matrix entries?

Affie:
@dehann do you perhaps know how the release testing of julia works? do they actually run you tests. 
Can it somehow cause a problem if the branch is set to not test on release branches?

dehann:
> do you perhaps know how the release testing of julia works? do they actually run you tests.

@Affie, currently Registry bots do not run individual package tests.  There was a time that JuliaCIBot (previous attobot world) would run all package as well as downstream dependencies as a matrix of tests.  The bot would flake out if your changes broke downstream packages on patch releases (ie forcing semver).  Will see if that comes back with new zepto system.

The current public testing only checks for Registry versioning type stuff.  Not sure, but they likely testing for security stuff privately somewhere too.  I'm not sure what is going on "behind the scenes" if there is something like that.  Although part of the assumption is  that public repos are heavily scrutinized by the public anyway.

> Can it somehow cause a problem if the branch is set to not test on release branches?

dont think so, and feel free to try!

docs/make.jl
```diff
@@ -1,5 +1,4 @@
 using Documenter
-using DataFrames
```
dehann: lol

dehann: thanks

PR #184, opened 2019-10-25
Closes #149 DFGVariable softtype parametric type

### LINKED ISSUE #149: DFGVariable softtype field has abstract type
Should DFG not export an abstract type with IIF InferenceType inheriting from it? I don't like that it does not have a type. see https://docs.julialang.org/en/v1/manual/performance-tips/#Avoid-fields-with-abstract-type-1
eg.`AbstractSoftType` then in IIF `InferenceType <: AbstractSoftType`
even `VariableNodeData{T<:Any}` will be better I think.

_Originally posted by @Affie in https://github.com/_render_node/MDI0OlB1bGxSZXF1ZXN0UmV2aWV3Q29tbWVudDMzMzQxNDkyMQ==/comments/review_comment_

Affie:
Set `VariableNodeData` to require an inference type - note that warnings are shown on older calls and this will cause result in many WARN's.


Affie:
**NOTE** So far this was just a way to handle depreciation and not the final implementation.
If it is ok to have breaking changes, without depreciation. ie.  `VariableNodeData() = error(""VariableNodeData() is depreciated please use VariableNodeData{T}() or VariableNodeData(softtype::T) where T <: InferenceVariable")`, it will make it a lot easier, but will have a big downstream impact.

dehann:
I vote skip the deprecation cycle since many other things are already in place to handle the transition.

My feeling is we are well prepared for the breaking change from DFG v0.4.x to DFG v0.5.0, since IIF v0.7.x is upper bound to only use DFG v0.4.x.  Therefore a clean transition to IIF v0.8.0 is possible.  Further, DFG currently has a single downstream dependency sream (IIF -> RoME -> Caesar -> Arena / RoMEPlotting) which similarly adhere to the compat structure.  So although we should do deprecation cycle through minor releases, we are at least doing half that and well prepared to quickly move on to IIF v0.8.0.  Further, the effort involved with supporting both with and without the template `T` is probably not worth it at this point.

dehann (approved):

GearsAD:
We'll merge this for 5.0 and then have to work through the cases with the warnings.

PR #199, opened 2019-11-18
Update DFGVariable.jl

GearsAD:


dehann:
Shall we aim for DFG v0.6 just incase it breaks downstream?  I think it will be good to template here -- same as we already do for Factors

dehann:
also see #237

GearsAD:
Part of the 4Q19/poc/v0_6 branch

PR #200, opened 2019-11-18
As per discussion - Type on solverDict

GearsAD:


dehann:
related to #237

dehann:
FYI -- what was the discussion on types for `solverDict` (assuming its an issue somewhere)?  My feeling is still that we should not do this since softtype is specific to a variable and not the solverDict elements inside a variable.  If you want to change a variable type, you need to build a whole new factor graph.

GearsAD:
Yes, if that's the case - "softtype is specific to a variable and not the solverDict elements inside a variable" - then my understanding is we merge #199 for v6.0 and this can be closed.

GearsAD:
Actually, while we're here, let's just line that up

PR #217, opened 2019-11-25
Moving isSolveInProgress to solver structures

GearsAD:
Not sure if this is truly a breaking change, but being a little conservative.


GearsAD:
Looks ok against IIF, actually. Will merge tomorrow.

Issue #221, opened 2019-11-27
isFactor true on variable


dehann:
Hi,  found a possible bug in that isFactor is returning true for variables:
```julia
julia> isFactor(fg,:x0)
true
```

This is using GraphsDFG.

dehann:
same is happening in reverse:
```julia
julia> isVariable(fg,:x0f1)
true
```

GearsAD:
Ah, this is because they both use exists, will make a small fix tomorrow.

dehann:
thanks!

PR #223, opened 2019-11-28
isVariable and isFactor

GearsAD:
I think that we need to standardize these functions in 0.6, but will be good for 0.5.


GearsAD:
Fix for #221

PR #224, opened 2019-11-28
tab = 4 spaces

Affie:


Affie:
I see my default change on src/DFGPlots/DFGPlots.jl slipped in.
I find it makes a better layout most of the time

GearsAD:
Thanks, will check out my Atom too, looks like I committed quite a bit of the tabbery.

PR #225, opened 2019-11-29
add setSolvable!

dehann:


dehann:
could we perhaps merge this please?

PR #227, opened 2019-11-30
Missing accessor methods for DFGVariable and DFGFactor 

dehann:


src/entities/DFGVariable.jl
```diff
@@ -159,12 +159,92 @@ DFGVariable(label::Symbol, softtype::T, _internalId::Int64 = 0) where {T <: Infe
...
+SkeletonDFGVariable(label::Symbol) = SkeletonDFGVariable(label, Symbol[])
+
 # Accessors
-label(v::DFGVariable) = v.label
-timestamp(v::DFGVariable) = v.timestamp
-tags(v::DFGVariable) = v.tags
-estimates(v::DFGVariable) = v.estimateDict
-estimate(v::DFGVariable, key::Symbol=:default) = haskey(v.estimateDict, key) ? v.estimateDict[key] : nothing
+
+const VariableDataLevel0 = Union{DFGVariable, DFGVariableSummary, SkeletonDFGVariable}
```
dehann: i was confused by Level0 vs Level2 in this context.  Level0 is right, but weird to get my head around how we should use the definitions here -- dispatch works!

dehann: i guess the way to think is that Level0 here is for functions that are available all the way up toLevel0 data, such as var.tags.  In contrast Level3 would allow many many more functions where the fields are not available in DFGVariableSkeleton etc.

PR #228, opened 2019-12-01
In-memory and File-based Bigdata stores

GearsAD:
FYI @Affie @dehann  - this changes the signatures for getBigDataEntries etc. to start following the standard CRUD pattern we spoke about.

It's not _completely_ finished so will put a few more changes in before tagging.

Note that every time you say BigData, you should airquote it :smile:


dehann:
xref #97

Issue #226, opened 2019-11-30
[FEATURE] maybe add timestamps to FactorDFG too


dehann:
not just variables

Affie:
For L1: Does a factor not already have the associated N variables’s timestamp with it?
For solver data, I think it might be handy as a factor type. Say you have an uncertain association to a moving object.

dehann:
also see #232

GearsAD:
Alrighty - merged!

Issue #182, opened 2019-10-23
Standardize graph CRUD API for update/merge/union


dehann:
To capture discussion and decision: want to make sure we have consistent update/merge/union behavior.  I think there was a renaming from update to merge at some point perhaps.  Something to the effect of (and free to disagree):
- `update` will update an existing key,
- `merge` will update or add depending on the key,
- `union` cannot update existing key but can add differing keys.

Strictly speaking union should return differing and both on collision, but we not ready to support all that functionality yet I think.

Affie:
What is the scope for this? Does it include addVariable/factor. It would be nice to have consistency everywhere.

dehann:
also entails question of:
https://github.com/JuliaRobotics/IncrementalInference.jl/blob/master/src/SubGraphFunctions.jl#L223
vs
https://github.com/JuliaRobotics/IncrementalInference.jl/blob/77e21ff4f145c83a3c7609eb3efcffbbb53e8717/src/CliqStateMachineUtils.jl#L642-L658

dehann:
Personally I think just make it work for DFG v0.5.0, but then we should think about 
> nice to have consistency everywhere

for DFG v0.6.0?

Affie:
ref: https://github.com/JuliaRobotics/IncrementalInference.jl/pull/395/files

GearsAD:
Sideline: So there's a larger discussion about how to stop users from using DFG.addVariable and start using IIF.addVariable for graphs that pack etc. It's led to a bit of confusion, trying to think of a good way to encourage users to use the IIF functions, and actually hide the DFG functions. Just planting the seed that it will affect this discussion.

Overall: These will be breaking changes as we would be changing signatures/behavior for addVariable/etc.. Let's do them for DFG 0.6.

As a philosophy, I'd like us to represent the underlying datastructure with no magic - e.g. most of our graph represents a dictionary - so implement standard behavior like a dictionary so that it feels familiar to users. Think of it as making sure users feel at home.

dehann:
> trying to think of a good way to encourage users to use the IIF functions,

- [ ] easiest is to not use reexport in IIF.
- [ ] would it work if there only abstract dispatch implementations in DFG, so when the user defines the graph type it forces specific functions.
[...]

PR #230, opened 2019-12-03
add timestamps to DFGFactor/Summary and closest functions

dehann:


dehann:
tests seem to fail due to compare with previously saved factor graph failing on a compare factor.

dehann:
also see #232

dehann:
oops, while using this `findVariableNearTimestamp(fg, ts, tags=[:POSE;])` JIT was unhappy about reducing an empty list.  Will investigate and fix!

src/CommonAccessors.jl
```diff
@@ -0,0 +1,108 @@
...
+  @warn "tags deprecated, use getTags instead"
+  getTags
+end
+
+"""
+$SIGNATURES
+
+Set the tags for a factor.
+"""
+setTags!(f::DataLevel0, tags::Vector{Symbol}) = f.tags = tags
```
Affie: This will not work on skeleton and depending on the final vote on #182 DFGVariable. 
I also think its a bad idea to allow memory pointers to be changed.

src/CommonAccessors.jl
```diff
@@ -0,0 +1,108 @@
...
+    @warn "timestamp deprecated, use getTimestamp instead"
+    getTimestamp(v)
+end
+
+"""
+$SIGNATURES
+
+Set the timestamp of a DFGFactor object.
+"""
+setTimestamp!(v::DataLevel1, ts::DateTime) = v.timestamp = ts
```
Affie: Also see #182. For me a timestamp is considered to be fixed to a variable and can not be changed. 
It is the same as the label in that sense. If you change the label its not the same variable any more.

src/CommonAccessors.jl
```diff
@@ -0,0 +1,108 @@
+# Common get and set methods
```
Affie: I kind of like the idea of the get methods by the types. It makes it easier to find. But doc strings will also help so I we move it here #233 becomes more important for me.

src/CommonAccessors.jl
dehann: > This will not work on skeleton

Got it thanks, so we can increase to `DataLevel1`?

> bad idea to allow memory pointers to be changed

So you'd suggest we rather use something like this?
```julia
function setTags!(f::DataLevel1, tags::Vector{Symbol})
  vtags = getTags(f)
  todelete = setdiff(vtags, tags)
  union!(vtags, tags)
  for td in todelete
    for ind in findall(x->x==td, vtags)
      delete!(vtags, ind..)
    end
  end
end
```

src/CommonAccessors.jl
dehann: I'm not so sure about that -- but if there is a strong performance reason then its worth it.  Otherwise you build a factor graph, realize you want to change something small and it becomes a major exercise.  This might be part of #182 for the usecase of changing immutable fields, we would need to create a new variable or factor object and replace the old one in the factor graph.  Will add a comment there.

src/CommonAccessors.jl
dehann: > get methods by the types

Sorry @Affie, what do you mean by that?  I think you did see #232 right?

Affie: Sorry wasn’t very clear. I mean for example  getTags(v::DFGvariable) in DFGVariable.jl by the struct. I thank thats where it came from. If all get methods are just get+fieldname then i guess it will be easy to find.

src/CommonAccessors.jl
Affie: Tags is in skeleton, but its not mutable. So you have to work with the original memory. Its not mutable for performance.

dehann: understood thanks, the suggested function here should work with the existing pointer -- i.e. with an immutable struct.

src/CommonAccessors.jl
dehann: thanks, I updated #232 accordingly

src/CommonAccessors.jl
Affie: with variable order in SkeletonFactor, I just did this:
```julia
resize!(factor._variableOrderSymbols, length(variableLabels))
factor._variableOrderSymbols .= variableLabels
```
so something like this seems a bit easier to read?
```julia
resize!(f.tags, length(tags))
f.tags .= tags
```

Affie: That's if you want to replace all of them.

dehann:
addresses #226

src/CommonAccessors.jl
GearsAD: Personally I like gets and sets (Java style) - think we all agree on that? A couple are not that way and they should be cleaned up.

src/CommonAccessors.jl
GearsAD: That means, at least for this PR, that summaries will fail on this call. This is a known issue until the immutable/mutable discussion is resolved.

src/CommonAccessors.jl
GearsAD: Just changed it out to that for now.

GearsAD:
Guys I'm just going to merge this and then battle out any changes in an effort to detangle some of this.

dehann:
Sounds good -- thanks and good luck!

PR #242, opened 2019-12-18
table of badges for readme

dehann:
Go here to see an example:
https://github.com/JuliaRobotics/DistributedFactorGraphs.jl/blob/readme/4Q19/tablulate/README.md


GearsAD:
@dehann thanks!

PR #253, opened 2020-01-08
save and load to gzip as default

dehann:


dehann:
FYI on motivation, just copying the last months results (~200Gb) is taking around 3 days due to the ~15million individual files.  tar.gz seems like a good compromise on speed, size, and file count.

dehann:
There were a few bugs, should be fixed and ready to go!

dehann:
I'd like to merge this in the next couple of days if there are no objections?

Issue #261, opened 2020-01-20
[BUG] subgraph functions not working for LightDFG


dehann:
```julia
julia> sfg = buildCliqSubgraph(fg, tree, :x1)
ERROR: getSubgraphAroundNode not implemented for LightDFG{SolverParams,DFGVariable,DFGFactor}
Stacktrace:
 [1] error(::String) at ./error.jl:33
 [2] #getSubgraphAroundNode#19(::Int64, ::typeof(getSubgraphAroundNode), ::LightDFG{SolverParams,DFGVariable,DFGFactor}, ::DFGVariable, ::Int64, ::Bool, ::GraphsDFG{SolverParams}) at /home/dehann/.julia/dev/DistributedFactorGraphs/src/services/AbstractDFG.jl:430
 [3] (::getfield(DistributedFactorGraphs, Symbol("#kw##getSubgraphAroundNode")))(::NamedTuple{(:solvable,),Tuple{Int64}}, ::typeof(getSubgraphAroundNode), ::LightDFG{SolverParams,DFGVariable,DFGFactor}, ::DFGVariable, ::Int64, ::Bool, ::GraphsDFG{SolverParams}) at ./none:0
 [4] #buildSubgraphFromLabels!#192(::GraphsDFG{SolverParams}, ::Int64, ::Array{Symbol,1}, ::typeof(buildSubgraphFromLabels!), ::LightDFG{SolverParams,DFGVariable,DFGFactor}, ::Array{Symbol,1}) at /home/dehann/.julia/dev/DistributedFactorGraphs/src/needsahome.jl:34
 [5] (::getfield(DistributedFactorGraphs, Symbol("#kw##buildSubgraphFromLabels!")))(::NamedTuple{(:subfg, :solvable, :allowedFactors),Tuple{GraphsDFG{SolverParams},Int64,Array{Symbol,1}}}, ::typeof(buildSubgraphFromLabels!), ::LightDFG{SolverParams,DFGVariable,DFGFactor}, ::Array{Symbol,1}) at ./none:0
 [6] buildCliqSubgraph(::LightDFG{SolverParams,DFGVariable,DFGFactor}, ::BayesTree, ::IncrementalInference.TreeClique, ::GraphsDFG{SolverParams}) at /home/dehann/.julia/dev/IncrementalInference/src/CliqStateMachineUtils.jl:1045
[...]

Issue #214, opened 2019-11-25
Mini cleanup on solvable for GetSubgraphAroundNode on LightGraphs


GearsAD:
Hi @Affie,

I added this https://github.com/JuliaRobotics/DistributedFactorGraphs.jl/blob/master/src/LightDFG/services/LightDFG.jl#L395 - but I think it will behave differently than the other drivers because it doesn't grow from 1 to `distance` neighbors. E.g. if in a graph x0 is solvable, x0x1f1 is not solvable, and x1 is solvable, asking for a subgraph around x0 distance=2,solvable=1 will return x1 even though x0x1f1 isn't solvable. 

I'm checking and adding a test for it, but do you mind taking a look at whether we can change it if needed?

GearsAD:
Branch with test that shows it doesn't maintain continuity - https://github.com/JuliaRobotics/DistributedFactorGraphs.jl/compare/bug/214_subgrapharoundnode

Affie:
Hi, looks like we have different approaches on getting the subgraph.
It will depend if it should be a connected sub graph or not. I get the entire sub-graph and then filter it. So it can have multiple fragments.

Related, why should the node you start with always be returned even if its not solvable?

GearsAD:
I'm worried about fragmented graphs, this has been a problem in the past. There's probably no hard and fast rule, but the idea of growing+filtering means that you will get a connected subgraph with all solvable.

For the first node, either can or can't return, but it's a boundary case - if you ask for a node that will be filtered then you're guaranteed to have a fragmented graph (strange user request).

GearsAD:
@dehann - do you have a strong preference either way? This is pretty much the only thing holding up 0.5.1.

Let there be a graph x0 -> x0x1f1 -> x1 -> x1x2f1 -> x2, and everything is SOLVABLE except x1. 
If I ask for getSubgraphAroundNode(..., x0x1f1, distance=2, solvable=1), then I could one of two results:
1. Solvable is not cascaded so I get a fragmented graph [x0 -> x0x1f1, x1x2f1 -> x2]
2. Solvable is cascaded, so I get [x0 -> x0x1f1]

No strong preference my side, I like the idea of connected graphs so I like 2, but that may just be because of past issues with fragmented graphs.

dehann:
yes my preference is a default with keyword argument to either selfCheckSolvable or nog.

fragmented graphs are hard to work with sometimes.

[...]

PR #262, opened 2020-01-20
Fix for 261

GearsAD:
`getSubgraphAroundNode` should allow any type of graph to be used for the `addToDFG` parameter.

Proposed fix for https://github.com/JuliaRobotics/DistributedFactorGraphs.jl/issues/261.

### DIFF
--- a/src/GraphsDFG/services/GraphsDFG.jl
+++ b/src/GraphsDFG/services/GraphsDFG.jl
@@ -334,7 +334,7 @@ Optionally provide a distance to specify the number of edges should be followed.
 Optionally provide an existing subgraph addToDFG, the extracted nodes will be copied into this graph. By default a new subgraph will be created.
 Note: By default orphaned factors (where the subgraph does not contain all the related variables) are not returned. Set includeOrphanFactors to return the orphans irrespective of whether the subgraph contains all the variables.
 """
-function getSubgraphAroundNode(dfg::GraphsDFG{P}, node::T, distance::Int64=1, includeOrphanFactors::Bool=false, addToDFG::G=GraphsDFG{P}(); solvable::Int=0)::GraphsDFG where {P <: AbstractParams, T <: DFGNode, G <: AbstractDFG}
+function getSubgraphAroundNode(dfg::GraphsDFG{P}, node::T, distance::Int64=1, includeOrphanFactors::Bool=false, addToDFG::G=GraphsDFG{P}(); solvable::Int=0)::AbstractDFG where {P <: AbstractParams, T <: DFGNode, G <: AbstractDFG}
     if !haskey(dfg.labelDict, node.label)
         error("Variable/factor with label '$(node.label)' does not exist in the factor graph")
     end
--- a/src/GraphsDFG/services/GraphsDFG.jl
+++ b/src/GraphsDFG/services/GraphsDFG.jl
@@ -334,7 +334,7 @@ Optionally provide a distance to specify the number of edges should be followed.
 Optionally provide an existing subgraph addToDFG, the extracted nodes will be copied into this graph. By default a new subgraph will be created.
 Note: By default orphaned factors (where the subgraph does not contain all the related variables) are not returned. Set includeOrphanFactors to return the orphans irrespective of whether the subgraph contains all the variables.
 """
-function getSubgraphAroundNode(dfg::GraphsDFG{P}, node::T, distance::Int64=1, includeOrphanFactors::Bool=false, addToDFG::GraphsDFG=GraphsDFG{P}(); solvable::Int=0)::GraphsDFG where {P <: AbstractParams, T <: DFGNode}
+function getSubgraphAroundNode(dfg::GraphsDFG{P}, node::T, distance::Int64=1, includeOrphanFactors::Bool=false, addToDFG::G=GraphsDFG{P}(); solvable::Int=0)::GraphsDFG where {P <: AbstractParams, T <: DFGNode, G <: AbstractDFG}
     if !haskey(dfg.labelDict, node.label)
         error("Variable/factor with label '$(node.label)' does not exist in the factor graph")
     end
--- a/src/LightDFG/services/LightDFG.jl
+++ b/src/LightDFG/services/LightDFG.jl
@@ -391,7 +391,7 @@ Optionally provide a distance to specify the number of edges should be followed.
 Optionally provide an existing subgraph addToDFG, the extracted nodes will be copied into this graph. By default a new subgraph will be created.
 Note: By default orphaned factors (where the subgraph does not contain all the related variables) are not returned. Set includeOrphanFactors to return the orphans irrespective of whether the subgraph contains all the variables.
 """
-function getSubgraphAroundNode(dfg::LightDFG{P,V,F}, node::DFGNode, distance::Int64=1, includeOrphanFactors::Bool=false, addToDFG::G=LightDFG{P,V,F}(); solvable::Int=0)::LightDFG where {P <: AbstractParams, V <: AbstractDFGVariable, F <: AbstractDFGFactor, G <: AbstractDFG}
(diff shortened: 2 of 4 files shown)


src/LightDFG/services/LightDFG.jl
```diff
@@ -347,41 +347,41 @@ function getNeighbors(dfg::LightDFG, label::Symbol; solvable::Int=0)::Vector{Sym
...
-
-            # Only if we have all of them should we add it (otherwise strange things may happen on evaluation)
-            if includeOrphanFactors || length(factVariables) == length(neigh_labels)
-                exists(destDFG, f.label) ? (@warn "_copyIntoGraph $(f.label) exists, ignoring") : addFactor!(destDFG, factVariables, deepcopy(f))
-            end
-        end
-    end
-    return nothing
-end
+# function _copyIntoGraph!(sourceDFG::LightDFG{<:AbstractParams, V, F}, destDFG::LightDFG{<:AbstractParams, V, F}, ns::Vector{Int}, includeOrphanFactors::Bool=false)::Nothing where {V <: AbstractDFGVariable, F <: AbstractDFGFactor}
```
dehann: we should remove this entirely

src/LightDFG/services/LightDFG.jl
```diff
@@ -391,7 +391,7 @@ Optionally provide a distance to specify the number of edges should be followed.
 Optionally provide an existing subgraph addToDFG, the extracted nodes will be copied into this graph. By default a new subgraph will be created.
 Note: By default orphaned factors (where the subgraph does not contain all the related variables) are not returned. Set includeOrphanFactors to return the orphans irrespective of whether the subgraph contains all the variables.
 """
-function getSubgraphAroundNode(dfg::LightDFG{P,V,F}, node::DFGNode, distance::Int64=1, includeOrphanFactors::Bool=false, addToDFG::LightDFG=LightDFG{P,V,F}(); solvable::Int=0)::LightDFG where {P <: AbstractParams, V <: AbstractDFGVariable, F <: AbstractDFGFactor}
+function getSubgraphAroundNode(dfg::LightDFG{P,V,F}, node::DFGNode, distance::Int64=1, includeOrphanFactors::Bool=false, addToDFG::G=LightDFG{P,V,F}(); solvable::Int=0)::AbstractDFG where {P <: AbstractParams, V <: AbstractDFGVariable, F <: AbstractDFGFactor, G <: AbstractDFG}
```
dehann: still curious why we have a specialized function here, must remove in time

src/GraphsDFG/services/GraphsDFG.jl
```diff
@@ -334,7 +334,7 @@ Optionally provide a distance to specify the number of edges should be followed.
 Optionally provide an existing subgraph addToDFG, the extracted nodes will be copied into this graph. By default a new subgraph will be created.
 Note: By default orphaned factors (where the subgraph does not contain all the related variables) are not returned. Set includeOrphanFactors to return the orphans irrespective of whether the subgraph contains all the variables.
 """
-function getSubgraphAroundNode(dfg::GraphsDFG{P}, node::T, distance::Int64=1, includeOrphanFactors::Bool=false, addToDFG::GraphsDFG=GraphsDFG{P}(); solvable::Int=0)::GraphsDFG where {P <: AbstractParams, T <: DFGNode}
+function getSubgraphAroundNode(dfg::GraphsDFG{P}, node::T, distance::Int64=1, includeOrphanFactors::Bool=false, addToDFG::G=GraphsDFG{P}(); solvable::Int=0)::AbstractDFG where {P <: AbstractParams, T <: DFGNode, G <: AbstractDFG}
```
Affie: With the new API, the verb would probably be something like deepcopy.

src/LightDFG/services/LightDFG.jl
Affie: I would like it to remain as a placeholder for an efficient LigthDFG to LightDFG copy.

src/needsahome.jl
```diff
@@ -21,11 +21,11 @@ Related
...
 getVariableIds, _copyIntoGraph!
 """
-function buildSubgraphFromLabels!(dfg::G,
+function buildSubgraphFromLabels!(dfg::GSRC,
                                   syms::Vector{Symbol};
-                                  subfg::AbstractDFG=(G <: InMemoryDFGTypes ? G : GraphsDFG)(params=getSolverParams(dfg)),
+                                  subfg::GDST=(GSRC <: InMemoryDFGTypes ? GSRC : GraphsDFG)(params=getSolverParams(dfg)),
                                   solvable::Int=0,
-                                  allowedFactors::Union{Nothing, Vector{Symbol}}=nothing  )::G where G <: AbstractDFG
+                                  allowedFactors::Union{Nothing, Vector{Symbol}}=nothing  )::GDST where {GSRC <: AbstractDFG, GDST <: AbstractDFG}
```
Affie: When this function was in IIF, the idea behind it was that in-memory graph types should not be converted between types and that other types would default to the graph defined in `const InMemDFGType`
I'm a bit worried that CGDFG might be broken.
It is probably a good idea to allow any` subfg <: InMemoryDFGTypes`, but this new definition does not make much sense. I will investigate.

src/LightDFG/services/LightDFG.jl
```diff
@@ -347,41 +347,41 @@ function getNeighbors(dfg::LightDFG, label::Symbol; solvable::Int=0)::Vector{Sym
...
-    return nothing
-end
+# function _copyIntoGraph!(sourceDFG::LightDFG{<:AbstractParams, V, F}, destDFG::LightDFG{<:AbstractParams, V, F}, ns::Vector{Int}, includeOrphanFactors::Bool=false)::Nothing where {V <: AbstractDFGVariable, F <: AbstractDFGFactor}
+#     #kan ek die in bulk copy, soos graph en dan nuwe map maak
+#     # Add all variables first,
+#     labels = [sourceDFG.g.labels[i] for i in ns]
+#
+#     for v in values(sourceDFG.g.variables)
+#         if v.label in labels
+#             exists(destDFG, v) ? (@warn "_copyIntoGraph $(v.label) exists, ignoring") : addVariable!(destDFG, deepcopy(v))
```
Affie: BTW, pending the decision on error behavior standardization, this might break `buildSubgraphFromLabels` completely as it adds duplicates and `add` with duplicates are supposed to error. That's why I just put the warning here and not an error.  I fixed the duplicates in my version of `getCliqueSubgraph` that was discarded.

src/LightDFG/services/LightDFG.jl
dehann: On philosophy, its about "cross-feeding" data from one DFG to another DFG and any whicha-way and abstracting away some of the technology concerns -- if its slightly slower, thats fine.  If it's much much slower with absolutely no recourse via a common abstract definition, then we can look at a performance specialization for high throughput cases.  See similar request there too: https://github.com/JuliaRobotics/DistributedFactorGraphs.jl/issues/214#issuecomment-558186644

Also separating out the design discussion:
#264

src/GraphsDFG/services/GraphsDFG.jl
dehann: Agreed, that is important for sure

src/needsahome.jl
dehann: Perhaps this fits in with the new talking point at #264

src/LightDFG/services/LightDFG.jl
dehann: You using a version that's working, but the common one is broken?

src/GraphsDFG/services/GraphsDFG.jl
GearsAD: With this you should be able to pass an destination graph and it should add to it, rather than assume a return of a GraphsDFG. 

Above, are we talking about renaming this to deepcopy?

Affie:
Closing as it is replaced by an abstract function in #263

src/GraphsDFG/services/GraphsDFG.jl
Affie: With new API `get` returns by reference if possible. So `build` or `deepcopy` fits better.
getSubgraph sounds like it should return a 'view'

Affie: I added `view` to the api discussion. Would be cool to implement something like `viewSubGraph` for the clique subgraphs.

PR #267, opened 2020-01-23
tags as a set, idea for consideration

Affie:
we can get the same behavior with accessors, but we cant guarantee that someone will not use v.tags directly. At least `union!([:a,:a], [:b]) == [:a, :b]` holds so I don't think it will matter much. I'm good with either, just thought I'd mention this way


dehann:
so there is also the direct `append!` approach used here:

https://github.com/JuliaRobotics/DistributedFactorGraphs.jl/blob/ed9fb94b7c649f01d9458a3ea9302a4cbccd72ca/test/interfaceTests.jl#L14-L16

GearsAD:
Sure, looks good!

Affie:
Jip, `append!` is part of the problem if its just a vector. You can cause this:
```julia
julia> x = [:a,:b,:c]
3-element Array{Symbol,1}:
 :a
 :b
 :c

julia> append!(x, [:a,:b,:c])
6-element Array{Symbol,1}:
 :a
 :b
 :c
 :a
 :b
 :c
```

Affie:
Also note that `append!` is not defined for `Set`

PR #268, opened 2020-01-26
add solveCount to DFGVariable

dehann:


dehann:
should rather move this down to VND instead

Issue #249, opened 2019-12-20
Incorrectly named function for adjacency matrix.


tonioteran:
I would opt to change the names of these functions, since they don't really return the adjacency matrix of the graph, nor do they return the incidence matrix.

They return the "boolean" version of the measurement Jacobian, which @dehann very cleverly calls it the first order Taylor expansion of the graph.

Probably `getMeasurementMatrix` would be better?

https://github.com/JuliaRobotics/DistributedFactorGraphs.jl/blob/05b6c1e8bc2d591a2506d7324a1f3a163a7dd791/src/services/AbstractDFG.jl#L539

https://github.com/JuliaRobotics/DistributedFactorGraphs.jl/blob/05b6c1e8bc2d591a2506d7324a1f3a163a7dd791/src/services/AbstractDFG.jl#L561

Affie:
I think its called a biadjacency matrix. Special case for a bipartite graph.

tonioteran:
I believe it cannot be an adjacency matrix, since it would have to be square, with rows and columns being the factor graph variables, and the ones representing edge existence. 

In this case we have the columns being the factor graph variables and the rows denoting the factors, with the ones expressing the variables involved in each one. Note we need not have a square matrix.

Here's a toy example of such a matrix.

   x1    x2    x3
 _                   _
|   1                  |   prior on x1
|          1           |   prior on x2
|                 1    |   prior on x3
|   1     1           |   odometry btwn x1 and x2
|_        1      1 _|   odometry btwn x2 and x3

More on this can be found on Kaess et al. IJRR paper on iSAM2 (Figure 3).

GearsAD:
Hi @tonioteran, yes I believe it's not a true adjacency matrix. If it leads to confusion we can change the name in v0.6. 

Btw cheat code for markdown table generator - https://www.tablesgenerator.com/markdown_tables


| x1 | x2 | x3 | desc                    |
|----|----|----|-------------------------|
| 1  |    |    | prior on x1             |
|    | 1  |    | prior on x2             |
|    |    | 1  | prior on x3             |
| 1  | 1  |    | odometry btwn x1 and x2 |
|    | 1  | 1  | odometry btwn x2 and x3 |

dehann:
I agree with Tonio to change the name, and would suggest we change it to:
- `getIncidenceMatrix[Sparse]`.  

[...]

PR #272, opened 2020-01-28
Interface tests pass except for a few compares and renamed to BiadjacencyMatrix

Affie:


src/LightDFG/services/LightDFG.jl
```diff
@@ -323,26 +323,10 @@ function getSubgraph(dfg::LightDFG{P,V,F}, variableFactorLabels::Vector{Symbol},
     return addToDFG
 end
 
-function getIncidenceMatrix(dfg::LightDFG; solvable::Int=0)::Matrix{Union{Nothing, Symbol}}
```
Affie: removed in favor of one abstract definition

src/services/AbstractDFG.jl
```diff
@@ -723,17 +724,17 @@ the symbol of the relating factor. The first row and first column are factor and
...
 """
     $(SIGNATURES)
 Get a matrix indicating relationships between variables and factors. Rows are
 all factors, columns are all variables, and each cell contains either nothing or
 the symbol of the relating factor. The first row and first column are factor and
 variable headings respectively.
+Note: rather use getBiadjacencyMatrix
 """
-function getIncidenceMatrix(dfg::AbstractDFG; solvable::Int=0)::Matrix{Union{Nothing, Symbol}}
+function getAdjacencyMatrixSymbols(dfg::AbstractDFG; solvable::Int=0)::Matrix{Union{Nothing, Symbol}}
```
Affie: We can keep it and rename it to something like this, but I don't think it should be used at all for calculations.

src/services/AbstractDFG.jl
```diff
@@ -750,25 +751,16 @@ function getIncidenceMatrix(dfg::AbstractDFG; solvable::Int=0)::Matrix{Union{Not
...
-"""
-    $(SIGNATURES)
-Get a matrix indicating relationships between variables and factors. Returned as
+Get a matrix indicating adjacency between variables and factors. Returned as
 a tuple: adjmat::SparseMatrixCSC{Int}, var_labels::Vector{Symbol)
 fac_labels::Vector{Symbol). Rows are the factors, columns are the variables,
 with the corresponding labels in fac_labels,var_labels.
 """
-function getIncidenceMatrixSparse(dfg::G; solvable::Int=0)::Tuple{SparseMatrixCSC, Vector{Symbol}, Vector{Symbol}} where G <: AbstractDFG
+function getBiadjacencyMatrix(dfg::G; solvable::Int=0)::Tuple{SparseMatrixCSC, Vector{Symbol}, Vector{Symbol}} where G <: AbstractDFG
```
Affie: Abstract function that can be overwritten for performance by drivers

src/services/AbstractDFG.jl
```diff
@@ -783,6 +775,10 @@ function getIncidenceMatrixSparse(dfg::G; solvable::Int=0)::Tuple{SparseMatrixCS
     return adjMat, varLabels, factLabels
 end
 
+function getAdjacencyMatrixSparse(dfg::G; solvable::Int=0)::Tuple{SparseMatrixCSC, Vector{Symbol}, Vector{Symbol}} where G <: AbstractDFG
```
Affie: I'm dropping Sparse for the default. Its silly to have a dense matrix. We can have getAdjacencyMatrixDense.

test/interfaceTests.jl
```diff
@@ -116,7 +117,7 @@ end
     global dfg,v1,v2,f1
     #TODO compare variable and factor
     @test getVariable(dfg, v1.label) == v1
-    @test getVariable(dfg, v2.label) != v1
+    @test_skip getVariable(dfg, v2.label) != v1
```
Affie: A few broken compares left in interface tests

dehann:
HI @Affie , we might need a round to figure out the right naming here.  Which information are you working from for Biadjacency -- think you mentioned Wikipedia?  We can resolve in #249 first, where near consensus was found for IncidenceMatrix.  Please comment there to resolve name first.

Affie:
I commented in #249 and https://github.com/JuliaRobotics/IncrementalInference.jl/issues/473, don't know what else there is to say.

dehann:
Cherry picking the separate commits perhaps?  Assuming the Adjacency renaming was in separate commits at least.

Affie:
Hi @dehann, I fixed some of the test and deprecated functions also. Its not worth the effort to split it.
It would be way easier just to "find replace" rename the function. 
Can you please go read through the 2 issues where this is discussed again and tell me what is wrong with `Biadjacency`. It looks like both Antonio and I don't feel `Incidence` is a good match. BTW, if we use Forney style FGs `Incidence` is correct, but half edges are weird.

src/services/AbstractDFG.jl
```diff
@@ -772,11 +773,10 @@ function getBiadjacencyMatrix(dfg::G; solvable::Int=0)::Tuple{SparseMatrixCSC, V
         factVars = getNeighbors(dfg, getFactor(dfg, factLabel), solvable=solvable)
         map(vLabel -> adjMat[fIndex,vDict[vLabel]] = 1, factVars)
     end
-    return adjMat, varLabels, factLabels
+    return (B=adjMat, varLabels=varLabels, facLabels=factLabels)
 end
 
-function getAdjacencyMatrixSparse(dfg::G; solvable::Int=0)::Tuple{SparseMatrixCSC, Vector{Symbol}, Vector{Symbol}} where G <: AbstractDFG
-    @warn "Deprecated function, please use getBiadjacencyMatrix as this will be removed in v0.6.1"
```
Affie: oops, accidentally deleted this line.

src/Common.jl
```diff
@@ -464,3 +463,29 @@ function hasTagsNeighbors(dfg::InMemoryDFGTypes,
   alltags = union( (ls(dfg, sym) .|> x->getTags(getNeiFnc(dfg,x)))... )
   length(filter(x->x in alltags, tags)) >= (matchAll ? length(tags) : 1)
 end
+
+
+
+#Natural Sorting
```
dehann: super super cool!

dehann:
@Affie go ahead, I tried to write a summary of it all here https://github.com/JuliaRobotics/IncrementalInference.jl/issues/473#issuecomment-581102453

Affie:
> Based on your example JuliaRobotics/IncrementalInference.jl#473 (comment) we should rather be getting

If that is your only reason then its not needed to do the transpose and I'd rather keep it the way its used in IIF now. 
The order in the Adjacency matrix returned by LigthGraphs only depend on the order the vertices are made, so this will also be a valid adjacency matrix for the same graph:

|         | :x1x2f1 | :x1 | :x2 |
|:-------:|:-------:|:---:|:---:|
| :x1x2f1 |    0    |  1  |  1  |
|   :x1   |    1    |  0  |  0  |
|   :x2   |    1    |  0  |  0  |

Most of them will look nothing like it as the variables and factors are mixed. I thought about implementing a structure for a bipartite graph type, but decided to rather stick with the standard SimpleGraphs structure in LightGraphs.

dehann:
Alright, don’t need the transpose — might be too much for general usage.  We should just write that in the function docs then, so future readers understand the reasoning.

src/services/AbstractDFG.jl
```diff
@@ -1,59 +1,89 @@
...
+"""
+    $(SIGNATURES)
+"""
+setSessionId!(dfg::AbstractDFG, sessionId::String) = dfg.sessionId = sessionId
+
+"""
+    $(SIGNATURES)
+"""
+#TODO don't know what error will be thrown if solverParams type does not mach the one in dfg
+setSolverParams!(dfg::AbstractDFG, solverParams::AbstractParams) = dfg.solverParams = solverParams
```
GearsAD: Expect a type-yelling error here, should be good for now though

src/services/AbstractDFG.jl
GearsAD: Agreed, if it's not used and the sparse is equivalent, let's remove it.

GearsAD:
Nice! Thanks for putting this all together

PR #278, opened 2020-02-05
Refactor comparison funcitons

Affie:
* Branch to look at comparisons in more detail, as they always seem to break


Affie:
The function that broke was an easy fix, but hard to find. It is also a lot of work to maintain. So I propose using generated functions for compare overloading ==

src/services/CompareUtils.jl
```diff
@@ -1,9 +1,40 @@
...
+TODO <:InferenceVariable: please someone confirm this, I'm not fully up to date why softtype was created.
+For now abstract `InferenceVariable`s are considered equal if they are the same type (except for labels, that I feel are deprecated)
+If your implentation has aditional properties such as `DynPose2` with `ut::Int64` (microsecond time) or support different manifolds
+implement compare if needed. softtype sounds a bit if it can be a trait, or in the future, rather a normal julia type.
+=#
+==(a::InferenceVariable,b::InferenceVariable) = typeof(a) == typeof(b)
+
+==(a::ConvolutionObject, b::ConvolutionObject) = typeof(a) == typeof(b)
+
+==(a::FunctorInferenceType, b::FunctorInferenceType) = typeof(a) == typeof(b)
```
dehann: just checking you know about `===`

dehann:
sounds good, merge when ready -- doesnt seem to have any breaking changes?

dehann:
also please add labels and milestones to this PR as appropriate :-)

Affie:
The big thing is that you have abstract compares that fall back to just checking types. The packages that inherit then have to implement compare if they want more.

src/services/CompareUtils.jl
Affie: Jip, but an object returned by a persistent driver will not be `===`

Affie:
Note, this is just to develop. When merging everything to master it will unfortunately  have breaking changes that we cannot avoid. Specifically the variable softtype parameter and API standardization changes in error behavior and return types.

dehann:
Sure, this is all for DFG v0.6, so breaking changes with deprecation warnings is fine (as much possible -- and more rigorous going forward).

PR #287, opened 2020-02-10
Consolidate test and some cleanup

Affie:


src/services/DFGVariable.jl
```diff
@@ -332,6 +334,7 @@ end
 
 Set solver data structure stored in a variable.
 """
+#TODO Repeated functionality?
```
Affie: this looks like its repeated in add/update

Affie:
@dehann  and @GearsAD, please have an overall look and tell me what you think of an idea along these lines. 
The main thing is I want to get rid of repeated code in tests that test for different thinks so I tried to split it in blocks (functions).
Summary types and interface tests are mostly consolidated. It fails because the cloud tests are still outdated on some of the new API, I will update it.

src/services/DFGVariable.jl
dehann: if so we should delete the duplicate

Affie: There are a few cases where functionality is duplicated. I deprecated some and aliased other. From slack Sam voted for deprecate for maintenance and I tend to agree.

PR #299, opened 2020-02-14
Refactor code structure, move without change

Affie:
Disclaimer: I don't take responsibility for you leaving me alone with the code.


Affie:
I merged as a review will not help much as all the code was moved around making tracking really hard.

Issue #305, opened 2020-02-16
FileDFG.saveDFG adds the extension even if the tar.gz extension is already there 


GearsAD:
```julia
# Example file will be test.tar.gz.tar.gz
saveDFG(fg, "/tmp/test.tar.gz");
```

dehann:
loadDFG should also be robust to either or, if for eg `/tmp/test` a tar file `/tmp/test.tar.gz`.

dehann:
think this was fixed by Sam

PR #307, opened 2020-02-16
Fixing tar file saving, comparator for Graphs.jl, and the FileDFG tests

GearsAD:
REF: #305

### DIFF
--- a/src/FileDFG/services/FileDFG.jl
+++ b/src/FileDFG/services/FileDFG.jl
@@ -11,18 +11,25 @@ dfg = initfg()
 # ... Add stuff to graph using either IIF or DFG:
 v1 = addVariable!(dfg, :a, ContinuousScalar, labels = [:POSE], solvable=0)
 # Now save it:
-saveDFG(dfg, "/tmp/saveDFG")
+saveDFG(dfg, "/tmp/saveDFG.tar.gz")
 ```
 """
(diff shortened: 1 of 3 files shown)


src/FileDFG/services/FileDFG.jl
```diff
@@ -11,18 +11,25 @@ dfg = initfg()
 # ... Add stuff to graph using either IIF or DFG:
 v1 = addVariable!(dfg, :a, ContinuousScalar, labels = [:POSE], solvable=0)
 # Now save it:
-saveDFG(dfg, "/tmp/saveDFG")
+saveDFG(dfg, "/tmp/saveDFG.tar.gz")
```
GearsAD: This is to just subtly push users to use the extension so we can smoothly remove the folder code in the future.

dehann: agree

GearsAD:
FYI @pvazteixeira

pvazteixeira:
thanks @GearsAD

Affie:
See #308

dehann:
Right, just avoid confusion -- we came to a decision how best to do compare functions (ie `==`and `isapprox`).  See #308

dehann:
Hi @GearsAD , thanks for the fix!

Issue #308, opened 2020-02-16
Does generated == replace the compare functions?


GearsAD:
Does this https://github.com/JuliaRobotics/DistributedFactorGraphs.jl/blob/master/src/services/CompareUtils.jl#L21-L28 replace the compare functions below?

Affie:
In my mind, no. I don't think we should change the julia definition of ==. Think compare can still be used for skipping and specialized test? I left compare untouched as it is what IIF uses for tests.
However, I like the skipfield idea, but maybe as a keyword argument. We can also consider isaprox for iif tests.

dehann:
so how about these two uses:
- A == B  ; and
- ==(A,B, skip=[:bob])

Affie:
I would then suggest API of == and `isapprox`/`≈`  with both ignore.
following julia on `isapprox` but with ignore added.

GearsAD:
:+1: for == ignore.

We just use the ignore for tests at the moment and I don't want us to break general use-cases - so it should be fine to just use the ==(..., ignore=[:adam]) there.

dehann:
so how about then also having:
- `≈(A, B, ignore=[:thing])`

dehann:
this may influence #285

dehann:
xref #307

Affie:
Also verbose flag for listing the fieldname it fails on.

dehann:
Keep in mind @Affie  built this with `@generated` macro, and I understand the fieldname only during verbose is easiest (no fancy messages).

dehann:
We need verbose to fix #293

Issue #293, opened 2020-02-12
Travis exceeds the maximum log length


Affie:
Output from the debug logger of these functions are displayed in travis:
https://github.com/JuliaRobotics/DistributedFactorGraphs.jl/blob/478b67f0e561407ba5525247fe98e3f8479d9241/src/services/CompareUtils.jl#L101

PR #309, opened 2020-02-18
Updated == and approx compares, verbose output

GearsAD:
How about this for #308 ? I haven't tested all cases it but it allows ignores and produces verbose output.

E.g.:
```julia
v1 = addVariable!(dfg, :a, ContinuousScalar, labels = [:POSE], solvable=0)
v2 = addVariable!(dfg, :b, ContinuousScalar, labels = [:LANDMARK], solvable=1)
v1 == v2
v1a = deepcopy(v1)
v1a._dfgNodeParams._internalId = 5


v1 ≈ v2
v1a = deepcopy(v1)
v1a._dfgNodeParams._internalId = 5
v1 ≈ v1a
```

Verbose output:
```
[ Info: DFGVariable{ContinuousScalar}.label: a != b
[ Info: DFGVariable{ContinuousScalar}.tags: Set(Symbol[:VARIABLE, :POSE]) != Set(Symbol[:LANDMARK, :VARIABLE])
[ Info: DFGNodeParams.solvable: 0 != 1
[ Info: DFGVariable{ContinuousScalar}._dfgNodeParams: DFGNodeParams(0, 0) != DFGNodeParams(1, 0)
[ Info: DFGNodeParams.solvable: 0 != 1
[ Info: DFGFactor{CommonConvWrapper{LinearConditional{Normal{Float64}}},Symbol}.label: abf1 != abf2
[ Info: DFGFactor{CommonConvWrapper{LinearConditional{Normal{Float64}}},Symbol}.timestamp: 2020-02-17T23:24:07.222 != 2020-02-17T23:24:07.242
```

### DIFF
--- a/src/services/CompareUtils.jl
+++ b/src/services/CompareUtils.jl
@@ -1,7 +1,8 @@
 ##==============================================================================
 ## (==)
 ##==============================================================================
-import Base.==
+import Base.==, Base.≈
+
 ## @generated compare
 # Reference https://github.com/JuliaLang/julia/issues/4648
 
@@ -23,11 +24,28 @@ const GeneratedCompareUnion = Union{MeanMaxPPE, VariableNodeData, DFGNodeParams,
                               GenericFunctionNodeData,
                               DFGFactor, DFGFactorSummary, SkeletonDFGFactor}
 
-@generated function ==(x::T, y::T) where T <: GeneratedCompareUnion
-    ignored = [:_internalId]
-    mapreduce(n -> :(x.$n == y.$n), (a,b)->:($a && $b), setdiff(fieldnames(x), ignored))
+function _compVerbose(type, prop, x, y, ignored)
+  @debug "Comparing: $type.$prop: $x ?= $y"
+  # Propagate the ignores if it's a GeneratedCompareUnion
+  if typeof(x) <: GeneratedCompareUnion
+      !(==(x,y, ignored)) && @info("$type.$prop: $x != $y")
+      return ==(x,y, ignored)
+  else
+      # Otherwise ignore ignored
+      !(==(x,y)) && @info("$type.$prop: $x != $y")
+      return ==(x,y)
+  end
+end
+function ==(x::T, y::T, ignored::Vector{Symbol}=Symbol[]) where T <: GeneratedCompareUnion
+    mapreduce(
+      n -> (_compVerbose(T, n, getfield(x,n), getfield(y,n), ignored)),
+      (a,b)-> (a && b),
+      setdiff(fieldnames(typeof(x)), ignored))
 end
 
+# Based these on the calls below, need to be confirmed
+≈(x::V, y::V) where V <: DFGVariable = ==(x,y, [:attributes, :solverDataDict, :_internalId, :createdTimestamp, :lastUpdatedTimestamp])
+≈(x::F, y::F) where F <: DFGFactor = ==(x,y, [:fnc, :attributes, :_variableOrderSymbols, :_internalId])
 
 ##==============================================================================
 ## Compare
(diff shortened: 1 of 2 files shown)


dehann:
Oh great thanks!  Were you able to use the defaults for ignore and verbose -- i.e. not use keywords to avoid @Affie 's earlier problem?

src/services/CompareUtils.jl
```diff
@@ -23,11 +24,28 @@ const GeneratedCompareUnion = Union{MeanMaxPPE, VariableNodeData, DFGNodeParams,
...
+function ==(x::T, y::T, ignored::Vector{Symbol}=Symbol[]) where T <: GeneratedCompareUnion
+    mapreduce(
+      n -> (_compVerbose(T, n, getfield(x,n), getfield(y,n), ignored)),
+      (a,b)-> (a && b),
+      setdiff(fieldnames(typeof(x)), ignored))
 end
 
+# Based these on the calls below, need to be confirmed
+≈(x::V, y::V) where V <: DFGVariable = ==(x,y, [:attributes, :solverDataDict, :_internalId, :createdTimestamp, :lastUpdatedTimestamp])
+≈(x::F, y::F) where F <: DFGFactor = ==(x,y, [:fnc, :attributes, :_variableOrderSymbols, :_internalId])
```
Affie: `≈` should call `≈` in the same way as `==`
`==` is the generic equality operator
`≈` `isapprox` is: Inexact equality comparison: true if `norm(x-y) <= max(atol, rtol*max(norm(x), norm(y)))`. 
It would be great if @dehann can write a `isapprox` for kde. We can dispatch on an `isparametric` trait to distinguish between them.

src/services/CompareUtils.jl
```diff
@@ -23,11 +24,28 @@ const GeneratedCompareUnion = Union{MeanMaxPPE, VariableNodeData, DFGNodeParams,
...
+  # Propagate the ignores if it's a GeneratedCompareUnion
+  if typeof(x) <: GeneratedCompareUnion
+      !(==(x,y, ignored)) && @info("$type.$prop: $x != $y")
+      return ==(x,y, ignored)
+  else
+      # Otherwise ignore ignored
+      !(==(x,y)) && @info("$type.$prop: $x != $y")
+      return ==(x,y)
+  end
+end
```
Affie: It would still cause #293.
What I have in mind is to have the generated `==` function for performance:
```julia
function ==(x::T, y::T; ignore::Vector{Symbol}=Symbol[]) where T <: GeneratedCompareUnion
  fields = setdiff(fieldnames(T), ignore)
  if @generated
    mapreduce(n -> :(x.$n == y.$n), (a,b)->:($a && $b), fields)
  else
    mapreduce(n -> getfield(x,n) == getfield(y,n), (a,b)-> a && b, fields)
  end
end
```
and then a `==(a,b,verbose; ignored=[])`
where verbose has levels something like:
1: only show type of compared on failure. ie "!= on val"
2: show type of compared and compact print on failure. "!= on value with x=[1,2, ...., 28,29,30], y=x=[1,2, ...., 28,29,30]
3: show everything.

src/services/CompareUtils.jl
dehann: JuliaRobotics/ApproxManifoldProducts.jl#31

dehann: id suggest accepting the error at this point and will be filled in once 31 is closed.

src/services/CompareUtils.jl
dehann: Again I suggest NOT using keywords but rather defaults:
`==(a,b,verbose, ignored=[])`

comma not semicolon

dehann: if we want to look the same as Base.==

dehann: another suggestion from slack was perhaps:
`==(a,b,verbose::Bool=false;ignore::Vector{Symbol}=[])`

I'd say as long as this does not conflict with Base and verbose + ignore are available.  Just do it and move on.

Affie:
continued in #480

PR #310, opened 2020-02-18
Add event::Threads.Condition to VND

### LINKED ISSUE #273: [FEAT] add Threads.Condition field to VND
Will allow user to "block" on a variable but immediately use the result after a notification (likely downsolve update), regardless of how far the total solver has progressed.

For example, image a graphinit operation is waiting on a previous variable somewhere.  If the previous variable is updated early in the downsolve on tree, then graph init can proceed without waiting for entire downsolve to complete first.

dehann:
fixes #273


dehann:
This needs some design first, see #273 for discussion

PR #334, opened 2020-02-28
optional GraphPlot in deps,

dehann:
cc @GearsAD -- in [extras] for optional packages


dehann:
ref JuliaGraphs/GraphPlot.jl#103

PR #347, opened 2020-03-03
better inner constructor

dehann:


src/entities/AbstractDFG.jl
```diff
@@ -27,8 +27,8 @@ The common node parameters for variables and factors.
 mutable struct DFGNodeParams
     solvable::Int
     _internalId::Int64
+    DFGNodeParams(s1=0,s2=0) = new(s1,s2)
```
Affie: I changed it to outer with a reference to from julia docs as why not to use inner unless its necessary, and now it back. There is no need for this. If you want this behavior, it should be something like:
```julia
DFGNodeParams(solvable=0) = DFGNodeParams(solvable, 0)
```

PR #354, opened 2020-03-04
Don't allow failures on IIF tests

Affie:
Test should always be run and kept in tact otherwise there is no point of testing as the coverage drops from 85% to 50%


dehann:
closes #354

Issue #338, opened 2020-02-29
Add specific functions for adding orphaned factors


Affie:
See #337 and related to #296

Affie:
We will only support orphaned factors if there is a need for it. Feel free to re-open or submit an new issue.

PR #371, opened 2020-03-30
Remove orphaned factors on deleteVariable!

### LINKED ISSUE #296: Remove variable neighbors (factors) by default on deleteVariable
If a variable is deleted from DFG object, then default is to delete surrounding factors also.  This can 
be avoided with a keyword argument.

If a factor is not deleted and a new variable is introduced, then a new factor must be created by one of the verbs and replace the current factor which has (by this time) a stale but immutable `_variableOrderList`.

One reason is that deleting and adding variables will not preserve the all important order.  Another reason is that deleting both the edges in the graph and variables from this list undermines the test where one compares the two lists against each other -- currently used for orphan test.  Third reason is when using the `multhypo` interface, then the number of variables is specific to that instance of the factor (as in all cases really) and therefore `_variableOrderList` must be immutable.

_Originally posted by @dehann in https://github.com/JuliaRobotics/DistributedFactorGraphs.jl/issues/20#issuecomment-585316742_

Affie:
close #296

Its a breaking change as the return type is changed and orphaned factors are removed from the graph, also see #338.

Orphaned factors are not currently supported and will be completely removed if they are not needed. (seems to be the case)


src/CloudGraphsDFG/services/CloudGraphsDFG.jl
```diff
@@ -262,22 +262,27 @@ function updateFactor!(dfg::CloudGraphsDFG, factor::DFGFactor)::DFGFactor
     return factor
 end
 
-function deleteVariable!(dfg::CloudGraphsDFG, label::Symbol)::DFGVariable
+function deleteVariable!(dfg::CloudGraphsDFG, label::Symbol)::Tuple{AbstractDFGVariable, Vector{<:AbstractDFGFactor}}
```
dehann, line 265: I have been removing all return types unless absolutely necessary (based on our previous conversation).

src/GraphsDFG/services/GraphsDFG.jl
```diff
@@ -161,14 +161,21 @@ function updateFactor!(dfg::GraphsDFG, factor::DFGFactor)::DFGFactor
     return factor
 end
 
-function deleteVariable!(dfg::GraphsDFG, label::Symbol)::DFGVariable
+function deleteVariable!(dfg::GraphsDFG, label::Symbol)::Tuple{AbstractDFGVariable, Vector{<:AbstractDFGFactor}}
```
dehann, line 164: similar

src/LightDFG/services/LightDFG.jl
```diff
@@ -119,14 +119,19 @@ function updateFactor!(dfg::LightDFG, factor::F)::F where F <: AbstractDFGFactor
     return factor
 end
 
-function deleteVariable!(dfg::LightDFG, label::Symbol)::AbstractDFGVariable
+function deleteVariable!(dfg::LightDFG, label::Symbol)::Tuple{AbstractDFGVariable, Vector{<:AbstractDFGFactor}}
```
dehann, line 122: similar

src/services/AbstractDFG.jl
```diff
@@ -226,7 +226,7 @@ end
     $(SIGNATURES)
 Delete a DFGVariable from the DFG using its label.
 """
-function deleteVariable!(dfg::G, label::Symbol)::AbstractDFGVariable where G <: AbstractDFG
+function deleteVariable!(dfg::AbstractDFG, label::Symbol)::Tuple{AbstractDFGVariable, Vector{<:AbstractDFGFactor}}
```
dehann, line 229: similar

src/services/AbstractDFG.jl
```diff
@@ -389,7 +389,7 @@ end
     $(SIGNATURES)
 Delete a referenced DFGVariable from the DFG.
 """
-function deleteVariable!(dfg::G, variable::V)::AbstractDFGVariable where {G <: AbstractDFG, V <: AbstractDFGVariable}
+function deleteVariable!(dfg::AbstractDFG, variable::AbstractDFGVariable)::Tuple{AbstractDFGVariable, Vector{<:AbstractDFGFactor}}
```
dehann, line 392: similar

PR #373, opened 2020-04-02
Refactor graph copy and subgraph functions

### LINKED ISSUE #300: Refactor graph copy functions
Look at draft of  `copyGraph` functions in attic/sandbox.jl
The idea is to have only one function to be called by functions such as `buildSubgraphFromLabels`

Will at least look at it for v0.6.0 and decide if it should be added.

Affie:
Close #300
Still to do:
- [x] add tests for `buildSubgraph`
- [x] deprecate `_copyIntoGraph` in favor of `copyGraph`/`deepcopyGraph`/`deepcopyGraph!`
- [x] deprecate `getSubgraph` in favor of `buildSubgraph`
- [x] deprecate `getSubgraphAroundNode` in favor of `buildSubgraph`


src/LightDFG/entities/LightDFG.jl
```diff
@@ -66,3 +66,14 @@ LightDFG(description::String,
          sessionData::Dict{Symbol, String},
          solverParams::AbstractParams) =
          LightDFG(FactorGraph{Int,DFGVariable,DFGFactor}(), description, userId, robotId, sessionId, userData, robotData, sessionData, Symbol[], solverParams)
+
+
+LightDFG{T,V,F}(description::String,
```
dehann, line 71: should this not perhaps be an inner constructor, or does it not matter?

dehann:
I think we can merge this, right?

GearsAD:
LGTM, just one thing - is there any reason (other than the original function signatures) that we combine variables and factors as input to `copyGraph!`, e.g. ` variableFactorLabels::Vector{Symbol};`?

This just results in the lists being split up first thing into the function. If there isn't a reason beyond the initial design, I think we should split it up to make the calls more efficient.

src/LightDFG/entities/LightDFG.jl
Affie: https://docs.julialang.org/en/v1/manual/constructors/#man-inner-constructor-methods-1

dehann: got it thanks!

GearsAD:
Btw, I meant that we can do that in a subsequent PR - I'll add that. This is good thanks.

GearsAD:
Oh I see you add it to the list already :) If you haven't started, let's save it for another PR

Affie:
Ok, will do separate PR. Just don't start using it in master yet.

PR #380, opened 2020-04-14
Feature ls2 close #332

### LINKED ISSUE #332: do we have a replacement for `ls2`?
Old use case, but might be nice to have something similar:

```julia
julia> ls2(fg, :x1)
[:x0; :x2; :l1]
```

Affie:
closes #332


dehann:
oh cool thanks!

PR #394, opened 2020-04-22
CGStructure updates

GearsAD:
Updates the functions required to persist the solver data via CGDFG

### DIFF
--- a/src/CloudGraphsDFG/entities/CGStructure.jl
+++ b/src/CloudGraphsDFG/entities/CGStructure.jl
@@ -10,8 +10,16 @@ mutable struct User <: AbstractCGNode
     id::Symbol
     name::String
     description::String
-    # labels::Vector{Symbol}
     data::Dict{Symbol, String}
+    createdTimestamp::String
+    lastUpdatedTimestamp::String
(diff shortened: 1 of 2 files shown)


src/CloudGraphsDFG/entities/CGStructure.jl
```diff
@@ -10,17 +10,34 @@ mutable struct User <: AbstractCGNode
     id::Symbol
     name::String
     description::String
-    # labels::Vector{Symbol}
     data::Dict{Symbol, String}
+    createdTimestamp::String
+    lastUpdatedTimestamp::String
+    User(id::Symbol,
```
Affie, line 16: https://docs.julialang.org/en/v1/manual/constructors/#man-inner-constructor-methods-1

Affie:
Do you only want this in v0.8.0?

src/CloudGraphsDFG/entities/CGStructure.jl
GearsAD: Huh, good note! Thanks, I'll move it outside.

GearsAD:
This is now merged to #417

PR #396, opened 2020-04-22
Cloud Tags SET Functions and tests

Affie:


src/CloudGraphsDFG/services/CloudGraphsDFG.jl
```diff
@@ -708,3 +708,126 @@ function setSolvable!(dfg::CloudGraphsDFG, sym::Symbol, solvable::Int)::Int
...
+                                            sym)
+
+    neo4jNode = Neo4j.getnode(dfg.neo4jInstance.graph, nodeId)
+
+    shouldStay = Symbol.([dfg.userId, dfg.robotId, dfg.sessionId]) ∪ [sym, :VARIABLE, :FACTOR]
+    tags = setdiff(Symbol.(getnodelabels(neo4jNode)), shouldStay)
+    # tags = Symbol.(getnodelabels(neo4jNode))
+
+    return removeTags!(dfg, sym, tags)
+end
```
Affie, line 833: @GearsAD If you have a few minutes, take a look at this. I'm thinking of not having the Neo4j labels repeated as tags in the variable data. But I implemented the variable tags version.

GearsAD: That's fair actually, we can subtract them, to leave the important ones like LANDMARK etc. I don't think there's a need for the user/robot/session labels. As long as they're on the neo4j node we're ok, that helps us search more efficiently, but the user doesn't need to see that. If I understand correctly, is that what you mean?

Issue #374, opened 2020-04-03
Dont use abstract return types


dehann:
Multiple dispatch is all about the calling types and not the return type.  Furthermore, trying to force a Abstract return type does not perform inheritance, but rather trying to find the necessary converters.  For example, this does not work:
```julia
   _       _ _(_)_     |  Documentation: https://docs.julialang.org
  (_)     | (_) (_)    |
   _ _   _| |_  __ _   |  Type "?" for help, "]?" for Pkg help.
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 1.3.1 (2019-12-30)
 _/ |\__'_|_|_|\__'_|  |  Official https://julialang.org/ release
|__/                   |

julia> abstract type MyAbstract end

julia> struct MyType <: MyAbstract
         a::Int
         end

julia> function ff(x::Int)::T where {T <: MyAbstract}
         MyType(x)
       end
ff (generic function with 1 method)

julia> ff(1)
ERROR: UndefVarError: T not defined
Stacktrace:
 [1] ff(::Int64) at ./REPL[3]:2
 [2] top-level scope at REPL[4]:1
```

Also, by defining a return type, Julia simply calls the convert and an assert on the return type.  It is not sensible to call `convert(::Abstract, ::Hardtype)` given a type stable function body.

cc @GearsAD , @Affie

Affie:
xref julia docs on return types: https://docs.julialang.org/en/v1/manual/functions/#Return-type-1

PR #402, opened 2020-04-25
add hasDataEntry, rm ReturnTypes

### LINKED ISSUE #401: add API `hasDataEntry`
useful API

## Entry or Element

Neither exist in the current Noun definitions on wiki.

dehann:
fix #401


dehann:
removing return types, see #374

Issue #415, opened 2020-04-28
Sentinel node behavior in CGDFG and matching InMemory


Affie:
We have to define how user/robot/session nodes are handled. 
My suggestion is to always force user/robot and auto generate session if it is not supplied. I previously used `string(uuid4())[1:6]`

user/robot can default to something like defaultUser/defaultRobot

PR #416, opened 2020-04-28
Update cgdfg tests and small bugfixes

Affie:
Some automatic session ids see #415 and tests @GearsAD (to help with the discussion)

### DIFF
--- a/test/consolInterfaceDev.jl
+++ b/test/consolInterfaceDev.jl
@@ -4,7 +4,7 @@ using Neo4j
 using DistributedFactorGraphs
 using Pkg
 using Dates
-
+using UUIDs
 # using IncrementalInference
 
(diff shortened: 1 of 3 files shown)


test/consolInterfaceDev.jl
```diff
@@ -4,7 +4,7 @@ using Neo4j
 using DistributedFactorGraphs
 using Pkg
 using Dates
-
+using UUIDs
```
dehann, line 7: not sure if we need to update Project.toml extras for UUID test dependency?

GearsAD: Should already be there?

Affie: We didn't include Dates with no problems, so I would say no.

PR #421, opened 2020-05-01
Validation during creation of DFG drivers

GearsAD:


Affie:
I'ts technically a breaking change, so I shifted milestone to v0.8.0

Issue #425, opened 2020-05-01
Abstract packing functions interface in DFG


Affie:
So far `decodePackedType` looks like the only one that has to be done.

GearsAD:
This is a good idea :+1:

Affie:
I missed `decodePackedTypeFunc` in the CGDFG struct, however it is not used. Abstract still seems like the way to go. I'll implement and then we can see how it works.

dehann:
HI @Affie , could you perhaps add some detail to the first comment about the issue here and #433 -- I'm not quite following?

Affie:
`decodePackedType` Is defined in IIF, but only used in one place in DFG. Its called in `Main` from DFG. 
I wanted to make a `decodePackedType` in DFG that IIF could add a method to, but it ended up not being possible with the current implementation. 
The only thing `decodePackedType` in DFG needs from IIF is the type to pack into, i.e. `CommonConvWrapper`. I didn’t want to get that from `Main` and it seemed reasonable to store it in the sover parameters.

dehann:
My understanding is `decodePackedType` has already been totally deprecated (but perhaps I'm mistaken)...  I actually did the work a while back to remove hard coded types/functions completely. 

> `decodePackedType` Is defined in IIF, but only used in one place in DFG

assume you mean this one:
https://github.com/JuliaRobotics/DistributedFactorGraphs.jl/blob/5d62dc7f7542ae63075f16ee4bf99a8d0fe413a1/src/services/Serialization.jl#L135

I thought the lifting was supposed to be done here?

https://github.com/JuliaRobotics/IncrementalInference.jl/blob/69266b45e8cc1a7bd80ae3e42535684219cd485c/src/DispatchPackedConversions.jl#L41-L59

and here:

https://github.com/JuliaRobotics/IncrementalInference.jl/blob/69266b45e8cc1a7bd80ae3e42535684219cd485c/src/DispatchPackedConversions.jl#L68

Although now that I'm looking closer it seems this second link approach was lost in translation and we are back at the hard coded method I had previously removed ... :-P  I'm trying to check and make sure but seems like it at this point.

## How To Fix

Replace this:
https://github.com/JuliaRobotics/DistributedFactorGraphs.jl/blob/5d62dc7f7542ae63075f16ee4bf99a8d0fe413a1/src/services/Serialization.jl#L135

with this a simple convert that will use this:
[...]

PR #433, opened 2020-05-03
decodePackedType with solverParams type, part of #425

Affie:


src/services/Serialization.jl
```diff
@@ -118,6 +118,17 @@ function packFactor(dfg::G, f::DFGFactor)::Dict{String, Any} where G <: Abstract
...
 
+
+function decodePackedType(::Type{T}, packeddata::GenericFunctionNodeData{PT,<:AbstractString}) where {T<:FactorOperationalMemory, PT}
+  # usrtyp = convert(FunctorInferenceType, packeddata.fnc)
+  # Also look at parentmodule
+  usrtyp = getfield(PT.name.module, Symbol(string(PT.name.name)[7:end]))
+  fulltype = DFG.FunctionNodeData{T{usrtyp}}
+  factordata = convert(fulltype, packeddata)
+  return factordata
+end
```
Affie, line 129: Moved here from IIF and signature changed to take the unpacking FactorOperationalMemory type

test/runtests.jl
```diff
@@ -85,7 +85,8 @@ end
 if get(ENV, "IIF_TEST", "") == "true"
 
     # Switch to our upstream test branch.
-    Pkg.add(PackageSpec(name="IncrementalInference", rev="upstream/dfg_integration_test"))
+    # Pkg.add(PackageSpec(name="IncrementalInference", rev="upstream/dfg_integration_test"))
+    Pkg.add(PackageSpec(name="IncrementalInference", rev="develop"))
```
Affie, line 89: temporary testing on develop again until v0.8.0 is ready

test/testBlocks.jl
```diff
@@ -61,6 +61,7 @@ struct TestCCW{T} <: FactorOperationalMemory where {T<:FunctorInferenceType}
 end
 Base.:(==)(a::TestCCW, b::TestCCW) = a.usrfnc! == b.usrfnc!
 
+DFG.getFactorOperationalMemoryType(par::NoSolverParams) = TestCCW
```
Affie, line 64: simulated type for testing

Affie:
@dehann, What do you think of using the SolverData for the `FactorOperationalMemory` type like this PR? 
From a generalization and testing perspective its a step in the right direction.

src/services/Serialization.jl
dehann: just breadcrum to remember to PR IIF for this change too

Affie: Yes, decodePackedType can be deleted from IIF if we do it this way.

src/CloudGraphsDFG/entities/CloudGraphsDFG.jl
```diff
@@ -11,9 +11,9 @@ mutable struct CloudGraphsDFG{T <: AbstractParams} <: AbstractDFG
     robotId::String
     sessionId::String
     description::String
-    encodePackedTypeFunc
-    getPackedTypeFunc
-    decodePackedTypeFunc
+    encodePackedTypeFunc #TODO Is this planed to be used or deprecated?
+    getPackedTypeFunc #TODO Is this planed to be used or deprecated?
+    decodePackedTypeFunc #TODO Is this planed to be used or deprecated?
```
dehann, line 16: My personal preference would be to use dispatch rather than hard wired function pointers.  I'm not sure if we'd be able to avoid the `"Packed"FactorType` approach which seems like a fair tradeoff.  If these fields are not used, then I'd say remove them ... although @GearsAD might have stronger recollection on why we had these and if they need to be kept.

src/services/AbstractDFG.jl
```diff
@@ -52,6 +52,13 @@ getAddHistory(dfg::AbstractDFG) = dfg.addHistory
 """
 getSolverParams(dfg::AbstractDFG) = dfg.solverParams
 
+"""
+    $(SIGNATURES)
+"""
+getFactorOperationalMemoryType(par::AbstractParams) = par.factorOperationalMemoryType
+
+getFactorOperationalMemoryType(dfg::AbstractDFG) = getFactorOperationalMemoryType(getSolverParams(dfg))
```
dehann, line 60: I think we might be missing each other here, we are not currently storing `ConvolutionalObject` so why would you need to store `FactorOperationalMemory`?  Should it not just be the same abstracts but better names and some of the IIF functionality moves up to DFG as needed?  My sense is quite against storing Type information in this way but im probably missing something?

dehann: okay so you did that in #422 , but then im missing what the idea is here?  I'm still not sure we should be storing Type information as currently suggested.

dehann: also see: https://github.com/JuliaRobotics/DistributedFactorGraphs.jl/pull/433#discussion_r419142906

dehann: right so, trying to do ... ? https://github.com/JuliaRobotics/DistributedFactorGraphs.jl/issues/425#issuecomment-623853861

Affie: Main idea is to make DFG not depend on any downstream functions to make testing and maintenance easier.

Affie: I'm open for other suggestions with the restraint that it is testable in DFG without needing IIF.

PR #436, opened 2020-05-04
Maintenance on GenericFunctionNodeData types

### LINKED ISSUE #274: GenericFunctionNodeData{T, S} T and S types
What type is T and can we inherit so its not any?
`fnc::T`

What is `frommodule`?, can it not be removed with new serialization in DFG or at least fixed to a `Symbol`
frommodule::S #Union{Symbol, AbstractString}

---

EDIT DF,

`frommodule` might be obsolete, lets plan to deprecate.

Affie:
part of  #274


dehann:
also see #441

PR #442, opened 2020-05-06
convenience buildSubgraph, fix #441

### LINKED ISSUE #441: REMOVE factorOperationalMemoryType from solverParams
See description of required work here: https://github.com/JuliaRobotics/DistributedFactorGraphs.jl/issues/425#issuecomment-624225618

issue is in response to having to merge #433 early and fix problems later.

## Background

More accurately, this field:
https://github.com/JuliaRobotics/DistributedFactorGraphs.jl/blob/d2f47c68eb86866786917176f2ed363a781a64b3/src/services/AbstractDFG.jl#L58

which in IIF case carries the value `CommonConvWrapper::UnionAll`

dehann:
closes #441

### DIFF
--- a/src/services/AbstractDFG.jl
+++ b/src/services/AbstractDFG.jl
@@ -54,10 +54,11 @@ getSolverParams(dfg::AbstractDFG) = dfg.solverParams
 
 """
     $(SIGNATURES)
-"""
-getFactorOperationalMemoryType(par::AbstractParams) = par.factorOperationalMemoryType
 
-getFactorOperationalMemoryType(dfg::AbstractDFG) = getFactorOperationalMemoryType(getSolverParams(dfg))
+Method must be overloaded by the user for Serialization to work.  E.g. IncrementalInference uses `CommonConvWrapper <: FactorOperationalMemory`.
+"""
+getFactorOperationalMemoryType(dummy) = error("Please define this function getFactorOperationalMemoryType for your usecase, e.g. = `IIF.CommonConvWrapper <: FactorOperationalMemory`")
+
 
 ##------------------------------------------------------------------------------
 ## Setters
--- a/src/services/AbstractDFG.jl
+++ b/src/services/AbstractDFG.jl
@@ -971,6 +971,14 @@ function buildSubgraph(::Type{G},
     return destDFG
 end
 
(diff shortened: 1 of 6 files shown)


dehann:
Others work, but trying to make CGDFG work with this: 
06c5602

test/consolInterfaceDev.jl
```diff
@@ -7,6 +7,12 @@ using Dates
 using UUIDs
 # using IncrementalInference
 
+## must overload the function
+import DistributedFactorGraphs: getFactorOperationalMemoryType
+# DF, dropping DFG. so that Main context is used and can be overridden by IIF.getFac...Type
+getFactorOperationalMemoryType(::CloudGraphsDFG) = TestCCW
```
Affie, line 13: Sorry, but this is a step back and breaks my testing consolidation effort

dehann: Think this is now resolved?  We using dispatch on ::AbstractParams rather than ::AbstractDFG as the compromise solution that avoids hard coding a variable somewhere.

src/services/AbstractDFG.jl
```diff
@@ -971,6 +972,14 @@ function buildSubgraph(::Type{G},
     return destDFG
 end
 
+# Perhaps IIF.InMemDFGType should perhaps be incorporated as a DFG value, but this has been much debated in the past and hence just left as breadcrum.
```
Affie, line 975: It doesn't really matter now since the only one is LightDFG. The main reason it was in iif was to help with testing and the transition between using GraphsDFG to LightDFG. We can make a DefaultDFGType const to use in case it ever changes in the future.

src/services/AbstractDFG.jl
```diff
@@ -971,6 +972,14 @@ function buildSubgraph(::Type{G},
...
+buildSubgraph(dfg::AbstractDFG,
+              variableFactorLabels::Vector{Symbol},
+              distance::Int=1;
+              solvable::Int=0,
+              sessionId::String = "",
+              kwargs...) = buildSubgraph(typeof(LightDFG(params=getSolverParams(dfg))), dfg, variableFactorLabels, distance, solvable=solvable, sessionId=sessionId, kwargs...)
```
Affie, line 981: Fix typeof(LightDFG(params=getSolverParams(dfg))) to not make an extra object that is not used

PR #443, opened 2020-05-06
fix #439

### LINKED ISSUE #439: Should this be Vector{Float64}
@dehann, should this perhaps be Vector{Float64}?
https://github.com/JuliaRobotics/DistributedFactorGraphs.jl/blob/5d62dc7f7542ae63075f16ee4bf99a8d0fe413a1/src/entities/DFGFactor.jl#L44

dehann:
fix #439


Affie:
I already did it somewhere, just wanted confirmation in #439 question. I'll sort PR's out

PR #447, opened 2020-05-06
Move GraphsDFG to attic and remove tests

Affie:
End of an era...


dehann:
Exciting stuff -- i think GraphDFG is near enough the medium term DFG design that it can be resurrected (in the unlikely event it's needed).  Thanks again!

All good, 100% go from my side!

dehann:
Oh, just remember that GraphsDFG has the slight regression of using Arrays not Dicts as was previously upgraded as such (to allow deletion of vertices etc.):
https://github.com/JuliaRobotics/IncrementalInference.jl/blob/fa574e9d06538512e76bc4561f2e1cd62f158cea/src/FactorGraphTypes.jl#L27

dehann:
Hi @Affie , i restarted the integration test on travis but no luck, perhaps you could look at that please and go ahead and merge when you are happy.. THanks

Affie:
Hi, Its a strange travis thing. It passes locally. Thats why I previously had DFG v0.7 also in deps.

PR #449, opened 2020-05-07
Also dispach on solverData type for rebuildFactorMetadata!

### LINKED ISSUE #425: Abstract packing functions interface in DFG

Affie:
Close #425

### DIFF
--- a/src/Deprecated.jl
+++ b/src/Deprecated.jl
@@ -11,6 +11,8 @@ include("../attic/GraphsDFG/GraphsDFG.jl")
 @reexport using .GraphsDFGs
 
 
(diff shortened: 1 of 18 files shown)


src/Deprecated.jl
```diff
@@ -11,6 +11,8 @@ include("../attic/GraphsDFG/GraphsDFG.jl")
 @reexport using .GraphsDFGs
 
 
+@deprecate loadDFG(source::String, iifModule::Module, dest::AbstractDFG) loadDFG!(dest, source)
```
Affie, line 14: To fit standard api

PR #476, opened 2020-05-22
add broadcasting

dehann:
I was busy here in `julia/base/abstractarray.jl`.  Something about `length(dfg) = 1` might be the problem, however, `add=+; add.(1,[2;3])` works fine while `length(::Int)==1` is always true... 

```julia
function copyto!(dest::AbstractArray, src)
    destiter = eachindex(dest)
    y = iterate(destiter)
    count = 0
    for x in src
        @show count += 1
        @show typeof(y)
        y === nothing &&
            throw(ArgumentError("destination has fewer elements than required"))
        dest[y[1]] = x
        y = iterate(destiter, y[2])
    end
    return dest
end
```

### DIFF
--- a/src/services/AbstractDFG.jl
+++ b/src/services/AbstractDFG.jl
@@ -1224,6 +1224,7 @@ import Base: length, iterate
 
 # to allow stuff like `getFactorType.(dfg, [:x1x2f1;:x10l3f2])`
 # not working properly yet -- trying to immitate `add=+; add.(3,[1;2])`
+# ie shortcut for getFactorType.([dfg;dfg], [:x1x2f1;:x10l3f2]) but using Julian interfaces properly
 # https://docs.julialang.org/en/v1/manual/interfaces/#
 length(::AbstractDFG) = 1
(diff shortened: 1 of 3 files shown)


dehann:
cc @Affie

src/services/AbstractDFG.jl
```diff
@@ -1216,3 +1216,14 @@ function getSummaryGraph(dfg::G)::LightDFG{NoSolverParams, DFGVariableSummary, D
...
+
+
+## features for broadcasting
+import Base: length, iterate
+
+# to allow stuff like `getFactorType.(dfg, [:x1x2f1;:x10l3f2])`
+# not working properly yet -- trying to immitate `add=+; add.(3,[1;2])`
+# https://docs.julialang.org/en/v1/manual/interfaces/#
+length(::AbstractDFG) = 1
+iterate(dfg::AbstractDFG, state::Int=0) = (dfg,state+1)
```
Affie: this would be needed for iterators and not for `getFactorType.(dfg, [:x1x2f1;:x10l3f2])`

PR #479, opened 2020-05-23
standardise to solverParams

Affie:
fix #475

Issue #472, opened 2020-05-19
ERROR: SystemError: opening file "..." Too many open files


Affie:
Files are not closed after reading.

Already fixed some

Issue #264, opened 2020-01-21
[DESIGN] "cross-feeding" vs "performance" -- ie dispatch vs specialized


dehann:
DFG is about having access to various graphing technologies in a way that allows info/data/etc to be easily transferred between technologies in a standard way.  For example:

```julia
merge!(CloudDFG, FileDFG)
merge!(FileDFG, CloudDFG)
merge!(CloudDFG, LightDFG)
merge!(GraphsDFG, LightDFG)
merge!(GraffDFG, CloudDFG) # seems excessive, but its about the philosophy 
# ...

update!(LigthDFG, CloudDFG)
update!(FileDFG, GraphDFG)
# ...

getSubGraph(::LightDFG, GraphDFG)
getSubGraph(::CloudDFG, LightDFG)
getSubGraph(::FileDFG, CloudDFG)
# ...
```

Originates from hard discussions on:
- #261 
- #214

dehann:
Decision is whether we want to support the kinds of "cross-feeding" functions I'm suggesting here -- or which subset?  Or, if any decisions are going to actively exclude any of the features suggested here?

dehann:
In reponse to a discussion in #261:

This is more speaking to the room:  I want to stress that DFG should not just be a clambering together of silo'ed drivers for graph technologies -- the most value comes from true "Distributed" nature in a variety of workloads that one user probably wont' see in their individual use-cases.  I fully acknowledge its hard to get right though.

dehann:
would you guys say the current design in DFG v0.8.0 achieves this -- so we can close?

Affie:
It’s definitely a step in that direction. I think it’s going to be a continuous effort.

dehann:
Cannot close until at least #507 is resolved

dehann:
https://github.com/JuliaRobotics/DistributedFactorGraphs.jl/issues/315#issuecomment-661027817

dehann:
xref:
- NavAbility/NavAbilitySDK.jl#298

dehann:
My sense is we very close to closing this issue?

dehann:
Oh wait -- there is big question here we have not touched on.  Which way round does the weakdeps go:
- NavAbilitySDK has a weakdep Caesar
- Caesar has a weakdep NavAbilitySDK

My kneejerk reaction is it should be the second Caesar has a weakdep

Affie:
> Which way round does the weakdeps go

[...]

PR #481, opened 2020-05-26
Mainly testing and maintenance

Affie:
see #471, #472, #476, #479, #478


src/Common.jl
```diff
@@ -1,10 +1,4 @@
 
-import Base: *
-
-
-# FIXME remove! is it really needed? This is type piracy
-*(a::Symbol, b::AbstractString)::Symbol = Symbol(string(a,b))
```
dehann, line 6: Yes, i use this -- not quite sure i see what is wrong with doing so?

src/DFGPlots/DFGPlots.jl
```diff
@@ -97,43 +97,50 @@ function gplot(dfg::LightDFG; keyargs...)
...
-            </script>
-            <script charset="utf-8">
-                $(read(GraphPlot.gadflyjs, String))
-            </script>
-            $(plotsvg)
-            </div>
-        """)
+#TODO decide if we want to overload show for display in juno, It's a bit annoying with development
+# function Base.show(io::IO, ::MIME"application/prs.juno.plotpane+html", dfg::AbstractDFG)
+function dfgplot(io::IO, ::MIME"application/prs.juno.plotpane+html", dfg::AbstractDFG)
```
dehann, line 102: Please do not overload show with plotting stuff.  Only if user wants to plot should that be done.

src/FileDFG/services/FileDFG.jl
```diff
@@ -127,24 +127,26 @@ function loadDFG!(dfgLoadInto::AbstractDFG, dst::String)
...
     @info "Loaded $(length(variables)) variables - $(map(v->v.label, variables))"
     @info "Inserting variables into graph..."
     # Adding variables
     map(v->addVariable!(dfgLoadInto, v), variables)
 
     for factorFile in factorFiles
-        io = open("$factorFolder/$factorFile")
-        packedData = JSON2.read(io, Dict{String, Any})
-        push!(factors, unpackFactor(dfgLoadInto, packedData))
+        open("$factorFolder/$factorFile") do io
```
dehann, line 141: oops, were these files not being closed?

src/services/AbstractDFG.jl
```diff
@@ -1,3 +1,13 @@
+##==============================================================================
+## AbstractDFG
+##==============================================================================
+##------------------------------------------------------------------------------
+## Broadcasting
+##------------------------------------------------------------------------------
+# to allow stuff like `getFactorType.(dfg, [:x1x2f1;:x10l3f2])`
+# https://docs.julialang.org/en/v1/manual/interfaces/#
+Base.Broadcast.broadcastable(dfg::AbstractDFG) = Ref(dfg)
```
dehann, line 9: oh thats easy, thanks!

src/Common.jl
Affie: https://docs.julialang.org/en/v1/manual/style-guide/index.html#Avoid-type-piracy-1

src/DFGPlots/DFGPlots.jl
Affie: Its how Images.jl and some other packages work, but I didn't like it so that's why its out. the result can be a handy summary though.

src/FileDFG/services/FileDFG.jl
Affie: Jip, so fixed the too many files open bug

src/services/DFGVariable.jl
```diff
@@ -46,7 +46,7 @@ getSofttype(dfg::AbstractDFG, lbl::Symbol, solvekey::Symbol=:default) = getSoftt
 
 Alias for [`getSofttype`](@ref).
 """
-getVariableType(params...) = getSofttype(params...)
+getVariableType(args...) = getSofttype(args...)
```
dehann, line 49: perhaps?

```julia
getVariableType(args...; kwargs...) = getSofttype(args...; kwargs...)
```

Affie: I'll have to look, This have been here for a while.

Affie: It doesn't take keyword arguments, so should be fine

dehann:
hi @GearsAD , FYI -- see temp license notice.  Johan is checking with minimal effort if CGDFG can function as submodule without distorting #264 (i.e. dispatch and no code duplication)

src/Common.jl
dehann: okay, leave it deleted then -- this case would probably not have caused too many problems but will toe the Julia line. thanks

src/services/AbstractDFG.jl
Affie: So you just have to tell julia dfg is a scalar.

src/Common.jl
Affie: Note, i'ts not used in DFG at all.

PR #483, opened 2020-05-27
Update Deprecated.jl * Symbol

dehann:
Hi @Affie ,  just a request that we please stick to proper deprecation cycles since there are downstream issues which cannot be untangled with consistent testing without doing deprecation rounds.  In this case tests in RoME are not able to jump to DFG v0.8.0 cleanly without this.

### DIFF
--- a/src/Deprecated.jl
+++ b/src/Deprecated.jl
@@ -7,6 +7,14 @@
 ## Remove in 0.9
 ##==============================================================================
 
+import Base: *
+
+# FIXME remove! is it really needed? This is type piracy
+function *(a::Symbol, b::AbstractString)::Symbol = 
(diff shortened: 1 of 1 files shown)


Affie:
Sorry tested only in DFG and IIF and couldn't find a place where it was used. I figured normal julia error would be easy enough.

Affie:
btw. this also works: `Symbol(:x0, "f1")` 
so the `*` is really not needed.

dehann:
> Sorry tested only in DFG and IIF and couldn't find a place where it was used. I figured normal julia error would be easy enough.

Yep, issue is Travis and version number are locked in -- hence semver :-)

PR #484, opened 2020-05-28
move cgdfg to a module in dfg

Affie:
~~- [ ] still needs some cleanup~~ Going to do this in separate PR with test improvements

### DIFF
--- a/src/CloudGraphsDFG/CloudGraphsDFG.jl
+++ b/src/CloudGraphsDFG/CloudGraphsDFG.jl
@@ -1,5 +1,93 @@
+module CloudGraphsDFGs
+
+using ...DistributedFactorGraphs
+
+import ...DistributedFactorGraphs:  setSolverParams!,
+                                    getFactor,
+                                    setDescription!,
+                                    # getLabelDict,
+                                    getUserData,
+                                    setUserData!,
+                                    getRobotData,
+                                    setRobotData!,
+                                    getSessionData,
+                                    setSessionData!,
+                                    addVariable!,
+                                    getVariable,
+                                    getAddHistory,
+                                    addFactor!,
+                                    getSolverParams,
+                                    exists,
+                                    isVariable,
+                                    isFactor,
+                                    getDescription,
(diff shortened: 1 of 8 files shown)


src/CloudGraphsDFG/CloudGraphsDFG.jl
```diff
@@ -1,6 +1,94 @@
...
+                                    getPPE,
+                                    addPPE!,
+                                    listPPEs,
+                                    deletePPE!,
+                                    updatePPE!,
+                                    packPPE,
+                                    unpackPPE,
+                                    getSolvable,
+                                    setSolvable!
+
```
Affie, line 76: Still need to confirm if everything is here.

src/DistributedFactorGraphs.jl
```diff
@@ -273,24 +274,28 @@ include("services/CompareUtils.jl")
...
         @info "Including Plots"
         include("DFGPlots/DFGPlots.jl")
         @reexport using .DFGPlots
     end
 
+    if get(ENV, "DFG_USE_CGDFG", "") == "true"
+        @info "Including CloudGraphsDFG"
+        @eval Base.include(DistributedFactorGraphs, joinpath(@__DIR__, "CloudGraphsDFG/CloudGraphsDFG.jl"))
+        @eval DistributedFactorGraphs @reexport using .CloudGraphsDFGs
+    end
```
Affie, line 297: @dehann How do we want to do this?
The other way is to do the check at compile time, but then you have to manually trigger the compile if you want to change it.

dehann: Not sure i understand the other way...?  But either way ENV variable is good/required, and its also fine to trigger new precompile if thats needed.

dehann:
looks good thanks @Affie

src/DistributedFactorGraphs.jl
Affie: Other way is not in `function __init__()`, so its pre-compiled.

Issue #492, opened 2020-06-02
Use tests for example code


Affie:
Use `jldoctest` `@example` code blocks to help avoid documentation examples from becoming outdated.

Affie:
see #493 for some examples of wow to do it.

PR #493, opened 2020-06-02
Docs examples

Affie:
ref #492


Affie:
Do not merge until tags are in. The example docs are dependant on IIF and it will not work until a new version is tagged.

dehann:
very cool, thanks!

PR #502, opened 2020-06-08
Cleaned up tests a bit and printVariable bugfix

Affie:


src/services/CustomPrinting.jl
```diff
@@ -98,7 +98,7 @@ end
 
 Display and return to console the user factor identified by tag name.
 """
-printFactor(dfg::AbstractDFG, sym::Symbol) = print(getFactor(dfg, sym))
+printFactor(dfg::AbstractDFG, sym::Symbol; kwargs...) = printFactor(getFactor(dfg, sym); kwargs...)
```
dehann, line 101: HI @Affie , does this default to `short=true` format?

dehann: I created a new format for factors that is easier to parse -- still had trouble with `@show fct` which fell back on the default field iteration.  Juno seemed to behave different from console REPL, but haven't check that out any further -- perhaps you fix is the answer for all those things?

PR #505, opened 2020-06-09
Update CGDFG 

dehann:


LICENSE
```diff
@@ -1,25 +1,212 @@
...
-MIT License for all components excluding CloudGraphDFG Driver.
+The source code in this repository is separated into two parts 
+that are licensed separately.  The CloudGraphDFG library/driver
+is licensed separately.  For more details see: 
+https://github.com/JuliaRobotics/DistributedFactorGraphs.jl/blob/develop/src/CloudGraphsDFG/LICENSE
 
-Copyright (c) 2017-2020 NavAbility
+The Caesar.jl ecosystem, including DistributedFactorGraphs.jl, is 
+not licensed for using, developing, or incorporating, in any 
+manner related to a public or private webservice or cloud system.
```
GearsAD, line 10: I would include the term commercial here, otherwise anyone that offers a free service using this code would be subject to this clause.

GearsAD: I think the Commons Clause would substitute in here quite well and it has the same effect, however it is widely used (https://commonsclause.com/)

Issue #514, opened 2020-06-27
Write variable+factor data in native formats


GearsAD:
That is:
- Timestamps as DateTime UTC
- Tags as array of strings
- Everything else as first-class properties

Affie:
Was this done in #517

PR #517, opened 2020-06-29
CloudGraphs DFG Updates

### LINKED ISSUE #513: Use the pose type to write formatted estimates to the DB

### LINKED ISSUE #515: Offload solverdata to separate nodes with fields as properties

### LINKED ISSUE #516: Offload BigData as separate nodes with fields as properties

GearsAD:
# Notes
Long overdue refactor/upgrade for CGDFG.

# TODO:
- [X] Add and fix unit tests
- [X] Variable and factor updates #514 
- [X] Using transactions for writing data #512 

# Requirements 
- Poses need to implement the `projectCartesian` function in RoME and IIF. 

# To Discuss
- I cleaned up getSofttype() (https://github.com/JuliaRobotics/DistributedFactorGraphs.jl/pull/517/files#diff-2fb9e13312751857b4b50f057b08bf34R40-R45) and then used that to shortcut how softtypes are retrieved from the DB. This assumes that softtype will always have a default constructor with no parameters. Any issues with that?

# Related PRs
- https://github.com/JuliaRobotics/RoME.jl/pull/304


src/CloudGraphsDFG/CloudGraphsDFG.jl
```diff
@@ -69,17 +70,16 @@ import ...DistributedFactorGraphs:  setSolverParams!,
...
+                                    setSolvable!,
+                                    getEstimateFields
 
 using Neo4j
 using Base64
 
 using DocStringExtensions
 using Requires
-using Dates
+using Dates, TimeZones
```
dehann, line 82: just a reminder to add compat in Project.toml

src/CloudGraphsDFG/services/CommonFunctions.jl
```diff
@@ -110,18 +112,22 @@ Will return all sessions because of 'node:Session'.
 If orderProperty is not ==, then 'order by n.{orderProperty} will be appended to the query'.
 So can make orderProperty = label or id.
 """
-function _getNeoNodesFromCyphonQuery(neo4jInstance::Neo4jInstance, matchCondition::String, orderProperty::String="")::Vector{Neo4j.Node}
+function _getNeoNodesFromCyphonQuery(neo4jInstance::Neo4jInstance, matchCondition::String, orderProperty::String=""; currentTransaction::Union{Nothing, Neo4j.Transaction}=nothing)::Vector{Neo4j.Node}
```
dehann, line 115: Also reminder that we decided against Julia function return types in general, here its `)::Vector{Neo4j.Node}`

src/CloudGraphsDFG/services/CommonFunctions.jl
```diff
@@ -176,28 +162,67 @@ function _tryGetNeoNodeIdFromNodeLabel(neo4jInstance::Neo4jInstance, userId::Str
...
 
 """
 $(SIGNATURES)
 
 Get the Neo4j labels for any node type.
 """
-function _getLabelsForType(dfg::CloudGraphsDFG, type::Type; parentKey::Union{Nothing, Symbol}=nothing)::Vector{String}
+function _getLabelsForType(dfg::CloudGraphsDFG,
+        type::Type;
+        parentKey::Union{Nothing, Symbol}=nothing)::Vector{String}
```
dehann, line 225: same

src/CloudGraphsDFG/services/CommonFunctions.jl
```diff
@@ -226,13 +253,108 @@ $(SIGNATURES)
...
+        tx(query)
+        result = commit(tx)
+    end
+    length(result.errors) > 0 && error(string(result.errors))
+    length(result.results[1]["data"]) != 1 && error("Cannot find subnode '$nodeKey' for variable '$variablekey'")
+    length(result.results[1]["data"][1]["row"]) != 1 && error("Cannot find subnode '$nodeKey' for variable '$variablekey'")
+    return result.results[1]["data"][1]["row"][1]
+end
+
+function _matchmergeVarSubnode!(
```
dehann, line 303: I think we should push for verbNoun compliance, i.e. `_matchmergeVariableSubnode!`.

Another cool thing is we have not formally defined the underscore `_verb` yet -- how about formal definition:
```
Underscore _verb* is a reserved convention for internal functions and variables 
that may not necessarily be useful to a user who is not interested in the internal 
developer aspects of the source code.  These functions, however, still follow
the general verbNoun convention as far possible.
```

dehann: cc @Affie

src/entities/DFGVariable.jl
```diff
@@ -32,6 +32,7 @@ mutable struct VariableNodeData{T<:InferenceVariable}
     dontmargin::Bool
     solveInProgress::Int
     solvedCount::Int
+    solverKey::Symbol
```
dehann, line 35: Changing type likely implies a new minor version series, shall we target DFG v0.9.0 for this then?

src/services/DFGVariable.jl
```diff
@@ -162,14 +162,14 @@ Since the `timestamp` field is not mutable `setTimestamp` returns a new variable
 Use [`updateVariable!`](@ref) on the returened variable to update it in the factor graph if needed. Alternatively use [`setTimestamp!`](@ref).
 See issue #315.
 """
-function setTimestamp(v::DFGVariable, ts::DateTime; verbose::Bool=true)
+function setTimestamp(v::DFGVariable, ts::ZonedDateTime; verbose::Bool=true)
```
dehann, line 165: WOW, this is a major breaking change!  This will make the entire ecosystem have to drop `DateTime` which is not reasonable I think -- not sure if we should resolve this with a conversion (think Julia will auto-promote):
```
convert(::Type{ZonedDateTime}, dt::DateTime)
```
so that when going into and out of Neo4j it does the conversion and then we can work on dispatch to support both cases `DateTime` and `ZonedDateTime`?

dehann: For v0.9.0 we would have to do all the deprecation warnings etc.

dehann (changes requested):

src/entities/DFGVariable.jl
Affie: I also just saw now that it is sometimes named `solverkey` and `solvekey`, 7 - 97

src/CloudGraphsDFG/services/CommonFunctions.jl
GearsAD: Lol, I'm a fan or strong typing everything so that it's more readable - is there a Julia-related reason why?

src/CloudGraphsDFG/services/CommonFunctions.jl
GearsAD: Agreed, will update it.  

Also - underscore functions are not exported.

src/entities/DFGVariable.jl
GearsAD: @Affie Yeah that was a big headache, I changed the ones I saw to `solverKey`. 
@dehann Yep, let's aim for 0.9.0

src/services/DFGVariable.jl
GearsAD: I'll internalize the CGDFG changes with ZonedDateTime, I was hoping it was a small set of changes, but it would go all the way to IIF. Rather, from now on DateTime is UTC and we'll handle ZonedDateTime inside CGDFG only. Changes on their way.

src/CloudGraphsDFG/CloudGraphsDFG.jl
GearsAD: Already in :+1:

src/services/DFGVariable.jl
dehann: two problems ecosystem wide, Timezone and Nanosecond accuracy.  Shall we do those together in future PRs?  Internalize in CGDFG for now and expand from there?

Affie: In case you haven't seen it, try `now(UTC)`, we can add that to all the default constructors. That way we always store DateTime in UTC, as you suggest, and then the functions on top of that can get the local time if needed.

That being said, It might be worthwhile to make the change to add Location. I think we can use `DateTime(zdt, Local)`
Maybe leave it as a separate PR?
I guess we can assume local time? 

@dehann, the `convert(::Type{ZonedDateTime}, dt::DateTime)` will also have to assume the TimeZone, but it is not lossless, so I would vote for using the constructor over converter.

Affie: see #523

src/CloudGraphsDFG/services/CommonFunctions.jl
Affie: The abstract return types gave me some issues as it calls `convert` and then `assert` automatically. I don't know about this case though.

src/services/DFGVariable.jl
GearsAD: :+1: Updated in IIF - https://github.com/JuliaRobotics/IncrementalInference.jl/pull/762/files

src/CloudGraphsDFG/services/CommonFunctions.jl
dehann: added definition to verbNoun documentation wiki

src/entities/DFGVariable.jl
```diff
@@ -157,11 +188,20 @@ struct MeanMaxPPE <: AbstractPointParametricEst
     mean::Vector{Float64}
     lastUpdatedTimestamp::DateTime
 end
+
 ##------------------------------------------------------------------------------
 ## Constructors
 
-MeanMaxPPE(solverKey::Symbol, suggested::Vector{Float64}, max::Vector{Float64},mean::Vector{Float64}) = MeanMaxPPE(solverKey, suggested, max, mean, now())
+MeanMaxPPE(solverKey::Symbol, suggested::Vector{Float64}, max::Vector{Float64}, mean::Vector{Float64}) = MeanMaxPPE(solverKey, suggested, max, mean, now(UTC))
```
Affie, line 195: I think we should make the timestamp consistently a ZonedDateTime

src/services/DFGVariable.jl
dehann: eg #582

PR #526, opened 2020-07-06
Rename BigData files without code changes

Affie:
I wasn't sure what to name files but didn't want to just use just Data.jl.

My suggestion would be that DataBlobs.jl (or another name) becomes a module.

I'm open for other names, I considered:
DataStores.jl
DFGDataStores.jl
DFGBlobs.jl
DFGDataBlobs.jl
DataEntriesBlobs.jl
DFGData.jl


dehann:
I'm happy with any one of those, and agree Data.jl is bit too bland.

PR #536, opened 2020-07-19
add GFND.nullhypo::Float64

dehann:
needed downstream for JuliaRobotics/IncrementalInference.jl#237

Not sure if this will effect CGDFG serialization....

@GearsAD, we can do DFG v0.10.0 for CGDFG and Data stuff if you will be busy for more than a few days in the current cycle?  No pressure at all, I'm just trying to sequence.  We will already have to punt @Affie 's IIF 459 downmsg consolidation work to IIF v0.15.0.  

IIF v0.13.0 landed yesterday and this DFG PR will result in IIF v0.14.0 to be placed directly after DFG v0.9.0 lands.  This is a breaking change to IIF.

### DIFF
--- a/src/entities/DFGFactor.jl
+++ b/src/entities/DFGFactor.jl
@@ -58,12 +58,12 @@ GenericFunctionNodeData{T}() where T =
 function GenericFunctionNodeData(eliminated::Bool,
                                  potentialused::Bool,
                                  edgeIDs::Vector{Int},
-                                 fnc,
+                                 fnc::T,
                                  multihypo::Vector{<:Real}=Float64[],
                                  certainhypo::Vector{Int}=Int[],
                                  nullhypo::Real=0,
-                                 solveInP::Int=0)
-    return GenericFunctionNodeData(eliminated, potentialused, edgeIDs, fnc, multihypo, certainhypo, nullhypo, solveInP)
+                                 solveInP::Int=0) where T
+    return GenericFunctionNodeData{T}(eliminated, potentialused, edgeIDs, fnc, multihypo, certainhypo, nullhypo, solveInP)
 end
 
 
--- a/src/entities/DFGFactor.jl
+++ b/src/entities/DFGFactor.jl
@@ -61,8 +61,9 @@ function GenericFunctionNodeData(eliminated::Bool,
                                  fnc,
                                  multihypo::Vector{<:Real}=Float64[],
                                  certainhypo::Vector{Int}=Int[],
-                                 nullhypo::Real=0)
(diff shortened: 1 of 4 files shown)


dehann:
FYI, these tests passed locally with `SKIP_CGDFG=true`

GearsAD:
Cool looks good. Lets see how far I get tomorrow with CGDFG and then make a call in evening? Don't want the CGDG changes to hold up IIF changes.

GearsAD:
CGDFG PR adds the solverkey to the VND, so this is going to conflict. Let's merge this and then I'll figure that out tomorrow.

GearsAD:
Wating for tests to pass...

dehann:
> Let's merge this and then I'll figure that out tomorrow.

Okay great thanks!

dehann:
> CGDFG PR adds the solverkey to the VND, so this is going to conflict.

This PR is only working with GFND so that should not interfere with VND, i think/hope.

dehann:
> Lets see how far I get tomorrow with CGDFG and then make a call in evening? Don't want the CGDG changes to hold up IIF changes.

It might just bump the minor version number one further.  If I could use v0.9.0 for these nullhypo and other low hanging changes, and the you take DFG v0.10.0 for CGDFG changes?

dehann:
hmmm, looks like the tests are failing here on CGDFG.  I'll dig in as soon I can, or perhaps @Affie is around on GMT-2 :-)

GearsAD (approved):

src/entities/DFGFactor.jl
```diff
@@ -39,23 +39,31 @@ mutable struct GenericFunctionNodeData{T<:Union{PackedInferenceType, FunctorInfe
...
-    multihypo::Vector{Float64} # FIXME likely to moved when GenericWrapParam is refactored #477
+    multihypo::Vector{Float64} # TODO re-evaluate after refactoring w #477
     certainhypo::Vector{Int}
+    nullhypo::Float64
     solveInProgress::Int
+
+    # TODO deprecate all these inner constructors at end of DFG v0.9.x (was added for GFND.nullhypo::Float64 breaking change)
+    GenericFunctionNodeData{T}() where T = new{T}()
+    GenericFunctionNodeData{T}(el,po,ed,fn,mu::Vector{<:Real},ce::Vector{Int},so::Int) where T = new{T}(el,po,ed,fn,mu,ce,0.0,so)
+    GenericFunctionNodeData{T}(el,po,ed,fn,mu::Vector{<:Real},ce::Vector{Int},nu::Real,so::Int) where T = new{T}(el,po,ed,fn,mu,ce,nu,so)
```
Affie, line 50: I know it will be deprecated, but why did you do it with inner constructors, I keep on removing unnecessary inner constructors:

PR #537, opened 2020-07-19
Update as per #533

### LINKED ISSUE #533: change Singleton / Pairwise to  AbstractPrior / AbstractRelativeFactor
Or something similar, we can discuss here

# Also see

- https://github.com/JuliaRobotics/IncrementalInference.jl/issues/783

GearsAD:
Closes #533

### DIFF
--- a/src/services/DFGFactor.jl
+++ b/src/services/DFGFactor.jl
@@ -135,7 +135,7 @@ Return `::Bool` on whether given factor `fc::Symbol` is a prior in factor graph 
 """
 function isPrior(dfg::G, fc::Symbol)::Bool where G <: AbstractDFG
   fco = getFactor(dfg, fc)
-  getFactorType(fco) isa AbstractPrior
+  getFactorType(fco) isa AbstractPrior || getFactorType(fco) isa FunctorSingleton
 end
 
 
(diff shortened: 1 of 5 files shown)


dehann:
skipped documntation test, so Travis did pass

src/services/DFGFactor.jl
```diff
@@ -135,7 +135,7 @@ Return `::Bool` on whether given factor `fc::Symbol` is a prior in factor graph
 """
 function isPrior(dfg::G, fc::Symbol)::Bool where G <: AbstractDFG
   fco = getFactor(dfg, fc)
-  getFactorType(fco) isa FunctorSingleton
+  getFactorType(fco) isa AbstractPrior || getFactorType(fco) isa FunctorSingleton
```
Affie, line 138: We should just remember to remove this with the deprecation.

dehann: sounds good, the deprecation warnings in DFG v0.9 should be sufficient.  Dropping old abstracts in DFG v0.10 is a good plan

GearsAD: I'm setting up this PR - there are a bunch of places they're still used, we just need to make sure v.10 won't break everything too badly.

PR #550, opened 2020-07-22
DataEntry and BlobStore Options

Affie:
Note: I gave functions new names to avoid having to deprecate everything. It's also temporary in a different file structure to make it easier.

There are basically 2 design concepts: BlobStoreEntry vs <: AbstractDataEntry
BlobStoreEntry has a generic entry with the blobstores for every type.
<: AbstractDataEntry has a custom type with no blobstore.

- BlobStoreEntry is the general data entry that goes with blob stores.
    - FolderStore
- Other AbstractDataEntry types (used without blobstores)
    - FileDataEntry
    - InMemoryDataEntry - save anything, persist nothing.
- DataBlob CRUD on <: AbstractDataEntry and AbstractDataStore
- Data CRUD on <: AbstractDataEntry and AbstractDataStore

TODO
- [x] Store stores in DFG? - @GearsAD Got this :+1:
- [x] Clean up
- [x] Remove old/deprecate
- [x] ~~Maybe choose one~~ EDIT: keep both for now, but focus on blobstore
- [x] ~~Some test and all methods~~ Expand tests and methods
- [x] Separate PR - perhaps Mongo blob CRUD - @GearsAD Yep separate is good. Opened issue to track: https://github.com/JuliaRobotics/DistributedFactorGraphs.jl/issues/552


src/Deprecated.jl
```diff
@@ -52,3 +52,118 @@ const AbstractBigDataEntry = AbstractDataEntry
...
+fg = initfg()
+addVariable!(fg, :x0, IIF.ContinuousScalar) # using IncrementalInference
+# FileDataStore (as example)
+datastore = FileDataStore(joinLogPath(fg,"datastore"))
+
+# now some data comes in
+mydata = Dict(:soundBite => randn(Float32, 10000), :meta => "something about lazy foxes and fences.")
+
+# store the data
+addDataEntry!( fg, :x0, datastore, :SOUND_DATA, "application/json", Vector{UInt8}(JSON2.write( mydata ))  )
```
dehann, line 137: this might be new API `addData!`?

src/Deprecated.jl
```diff
@@ -52,3 +52,118 @@ const AbstractBigDataEntry = AbstractDataEntry
...
+
+# unpack data to original format (this must be done by the user)
+@show entry.mimeType # "applicatio/json"
+userFormat = JSON2.read(IOBuffer(rawData))
+```
+
+Related
+
+addDataEntry!, addData!, fetchData, fetchDataEntryElement
+"""
```
dehann, line 150: if we deprecating this, what is the direct replacement for this documented example in the new API -- specifically the combo?
```julia
addDataEntry!( fg, :x0, datastore, :SOUND_DATA, "application/json", Vector{UInt8}(JSON2.write( mydata ))  )
entry, rawData = fetchData(fg, :x0, datastore, :SOUND_DATA)
```

Affie: I don't know what default to use yet. For persistence: the `FolderStore`, for quick and easy (not persisted or copied to the cloud) I like the in-memory data entry.

We currently have:
```julia
dataset1 = rand(UInt8, 1000)

ade,adb = addData!(InMemoryDataEntry, dfg, :x1, :random, dataset1)
#or
ade,adb = addData!(FileDataEntry, dfg, :x1, :random, "/tmp/dfgFilestore", dataset1)
#or
ds = FolderStore{Vector{UInt8}}(:filestore, "/tmp/dfgFilestore")
addBlobStore!(dfg, ds)
ade,adb = addData!(dfg, :filestore, :x1, :random, dataset1)
```
The docs are a bit behind and we need good examples. The main purpose of PR was to have a clean API.

Affie: Direct replacement should be (if the store is not saved in the DFG):
```julia
blobstore = FolderStore{Vector{UInt8}}(:filestore, "/tmp/dfgFilestore")
addData!(fg, blobstore, :x0, :SOUND_DATA, Vector{UInt8}(JSON2.write( mydata )), mimeType="application/json")

rawData, entry = getData(fg, blobstore, :x0, :SOUND_DATA)
```

Affie: Sorry for changing the parameter order. The DFG and BlobStore are mutated so they should go first and there were ambiguities with dispatch.

The idea is to only need the CRUD verbs `getData/addData!/updateData!/deleteData!` and not have other verbs such as fetch.

Affie: CRUD API
- Verbs: get,add,update,delete
- Nouns: Data, DataBlob, DataEntry

dehann: Ah, gotcha thanks!  This helps

dehann: > Sorry for changing the parameter order.

Totally fine, I'm just trying to follow

Issue #554, opened 2020-07-26
Evaluation into the closed module CGDFG breaks (don't do this)


dehann:
## Suggested Solution

Here is what was suggested by Tim Holy for a similar error:
- https://github.com/JuliaGraphics/ColorTypes.jl/pull/178#issuecomment-614902798
- JuliaLang/julia#35501

-------

## Original Error

Oops, this is new.  Just noticed on testing with IIF (this is the DEV MASTERS Suite)
- https://travis-ci.org/github/JuliaRobotics/IncrementalInference.jl/jobs/711863747
- https://travis-ci.org/github/JuliaRobotics/IncrementalInference.jl/jobs/711865177

EDIT, note #556 has since been added

```julia
ENV["DFG_USE_CGDFG"]=true

using IncrementalInference
...
[ Info: Including CloudGraphsDFG

ERROR: LoadError: InitError: Evaluation into the closed module `DistributedFactorGraphs` breaks incremental compilation because the side effects will not be permanent. This is likely due to some other module mutating `DistributedFactorGraphs` with `include` during precompilation - don't do this.
Stacktrace:
 [1] include(::Function, ::Module, ::String) at ./Base.jl:380
 [2] include at ./Base.jl:368 [inlined]
 [3] __init__() at /home/travis/.julia/packages/DistributedFactorGraphs/r6Be7/src/DistributedFactorGraphs.jl:295
 [4] _include_from_serialized(::String, ::Array{Any,1}) at ./loading.jl:697
 [5] _require_from_serialized(::String) at ./loading.jl:749
 [6] _require(::Base.PkgId) at ./loading.jl:1040
 [7] require(::Base.PkgId) at ./loading.jl:928
 [8] require(::Module, ::Symbol) at ./loading.jl:923
 [9] include(::Function, ::Module, ::String) at ./Base.jl:380
 [10] include(::Module, ::String) at ./Base.jl:368
 [11] top-level scope at none:2
 [12] eval at ./boot.jl:331 [inlined]
 [13] eval(::Expr) at ./client.jl:467
 [14] top-level scope at ./none:3
during initialization of module DistributedFactorGraphs
in expression starting at /home/travis/build/JuliaRobotics/IncrementalInference.jl/src/IncrementalInference.jl:15
[...]

PR #556, opened 2020-07-26
update info printout cgdfg

### LINKED ISSUE #555: Reintroduce if DFG_USE_CGDFG Environment Variable
This was temporarily removed in #517 but is required for licensing reasons.
https://github.com/JuliaRobotics/DistributedFactorGraphs.jl/blob/c956be5aba890e5a3ef93292a52622b25cd08a45/src/DistributedFactorGraphs.jl#L293

dehann:
Hi Sam, if you wanted to review when you get a chance -- didn't want to bug you on a busy week.  Also see #555

### DIFF
--- a/src/DistributedFactorGraphs.jl
+++ b/src/DistributedFactorGraphs.jl
@@ -290,11 +290,11 @@ function __init__()
         @reexport using .DFGPlots
     end
 
-    # if get(ENV, "DFG_USE_CGDFG", "") == "true"
+    if get(ENV, "DFG_USE_CGDFG", "") == "true"
         @info "Detected ENV[\"DFG_USE_CGDFG\"]: Including optional CloudGraphsDFG (LGPL) Driver"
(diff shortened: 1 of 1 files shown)


GearsAD (approved):

dehann:
thanks!

src/DistributedFactorGraphs.jl
```diff
@@ -291,7 +291,7 @@ function __init__()
     end
 
     # if get(ENV, "DFG_USE_CGDFG", "") == "true"
```
GearsAD, line 293: Good point, think we want the if statement as well?

dehann: that will close #555 too, we can add whenever is good.  Just helping review code while I chew on other stuff.  You should hide and focus on big items for this week.  We can catch up after.

I will add and let you merge when you get a moment.

dehann:
closes #555

src/DistributedFactorGraphs.jl
dehann: ha, thanks!

dehann:
FYI, I also found suggestions to fix #554, see updated first comment in 554

PR #567, opened 2020-07-29
Precompile CGDFG based on ENV variable

Affie:
Temporary fix for #554
This is not really a good idea. It looks like new updates to Pkg might provide a better solution.
Note that ENV variable DFG_USE_CGDFG should be seen as a permanent feature setting as CGDF will be precompiled into the cache.

### DIFF
--- a/src/DistributedFactorGraphs.jl
+++ b/src/DistributedFactorGraphs.jl
@@ -289,18 +289,31 @@ include("Common.jl")
 # Data Blob extensions
 include("DataBlobs/DataBlobs.jl")
 
+if get(ENV, "DFG_USE_CGDFG", "") == "true"
+    @info "Detected ENV[\"DFG_USE_CGDFG\"]: Including optional CloudGraphsDFG (LGPL) Driver"
+    include("CloudGraphsDFG/CloudGraphsDFG.jl")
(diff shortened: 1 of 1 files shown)


GearsAD:
I think this is good - if you want to disable/enable this, you need to remove and reload DFG with that flag set. 

We can revisit when Pkg functionality exists.

dehann:
I think there is a trick that Tim Holy suggested (as usual) -- somewhere out there in the ecosystem comments:
```julia
ENV = ENV["DFG_USE_CGDFG"]
using Pkg
Pkg.build("DistributedFactorGraphs")
using DistributedFactorGraphs
```

Affie:
```julia
julia> ENV["DFG_USE_CGDFG"]="true"
julia> Base.compilecache(Base.PkgId(DistributedFactorGraphs))
[ Info: Precompiling DistributedFactorGraphs [b5cc3c7e-6572-11e9-2517-99fb8daf2f04]
[ Info: Detected ENV["DFG_USE_CGDFG"]: Including optional CloudGraphsDFG (LGPL) Driver
```

Affie:
`Base.compilecache(Base.PkgId(DistributedFactorGraphs))` seems to not work with Revise

PR #570, opened 2020-08-03
Simple copy functions

GearsAD:
This work for simple copy functions that we can use in IIF?

### DIFF
--- a/src/entities/DFGVariable.jl
+++ b/src/entities/DFGVariable.jl
@@ -87,6 +87,28 @@ VariableNodeData(val::Array{Float64,2},
 VariableNodeData(softtype::T; solverKey::Symbol=:default) where T <: InferenceVariable =
     VariableNodeData{T}(zeros(1,1), zeros(1,1), Symbol[], Int[], 0, false, :NOTHING, Symbol[], softtype, false, 0.0, false, false, 0, 0, solverKey)
 
+## Copying
+function copy(vnd::VariableNodeData; solverKey::Union{Nothing,Symbol}=nothing)
+    return VariableNodeData(
+        deepcopy(vnd.val),
+        deepcopy(vnd.bw),
+        deepcopy(vnd.BayesNetOutVertIDs),
+        deepcopy(vnd.dimIDs),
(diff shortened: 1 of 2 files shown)


src/entities/DFGVariable.jl
```diff
@@ -87,6 +87,28 @@ VariableNodeData(val::Array{Float64,2},
 VariableNodeData(softtype::T; solverKey::Symbol=:default) where T <: InferenceVariable =
     VariableNodeData{T}(zeros(1,1), zeros(1,1), Symbol[], Int[], 0, false, :NOTHING, Symbol[], softtype, false, 0.0, false, false, 0, 0, solverKey)
 
+## Copying
+function copy(vnd::VariableNodeData; solverKey::Union{Nothing,Symbol}=nothing)
```
Affie, line 91: Maybe we shouldn't overwrite expected behaviour from base copy with deepcopy?
I expect `copy` to return a shallow copy and deepcopy for this.

Personally I don't think we should provide too many functions. Every function has to be documented and maintained.
The user can rather just do 
```juila
dcvnd = deepcopy(vnd)
dcvnd.selverKey = newKey
```

Is the function we are trying to replace `updateSolverData!(dfg, label, vnd, newsolvekey)`?
Where `vnd.solvekey != newsolvekey`.
Perhaps something like (although IIF needs to be updated slightly for it):
```julia
function deepcopySolverData!(dfg::AbstractDFG, label::Symbol, srcKey::Sybol, dstKey::Symbol)
    vnd = deepcopy(getSolverData!(dfg, label, srcKey))
    vnd.solverKey = dstKey
    updateSolverData!(dfg, vnd)
end  
```
Now that I'm thinking it through. `update` is maybe not the correct function to use. Its adding new solverdata so maybe:
```julia
function addSolverData!(dfg::AbstractDFG, label::Symbol, vnd, solverKey::Symbol)
    vnd.solverKey = dstKey
    addSolverData!(dfg, vnd)
end  
```

GearsAD: Cool, that works, will close this then.

PR #575, opened 2020-08-06
Smalldata update

GearsAD:


Affie:
@GearsAD, I'm branching of the previous commit and will add the in-memory crud. So will replace this.

PR #583, opened 2020-08-09
Macro to create a new <:InferenceVariable

Affie:


dehann:
Johan, you should merge when ready!

PR #585, opened 2020-08-10
CGDFG blobs as nodes

GearsAD:
External DataEntry nodes in CGDFG.

### DIFF
--- a/src/DataBlobs/services/BlobStores.jl
+++ b/src/DataBlobs/services/BlobStores.jl
@@ -160,9 +160,9 @@ function addDataBlob!(store::FolderStore{T}, entry::BlobStoreEntry, data::T) whe
     blobfilename = joinpath(store.folder,"$(entry.id).dat")
     entryfilename = joinpath(store.folder,"$(entry.id).json")
     if isfile(blobfilename)
-        error("Key '$(id)' blob already exists.")
+        error("Key '$(entry.id)' blob already exists.")
     elseif isfile(entryfilename)
-        error("Key '$(id)' entry already exists, but no blob.")
+        error("Key '$(entry.id)' entry already exists, but no blob.")
     else
         open(blobfilename, "w") do f
             write(f, data)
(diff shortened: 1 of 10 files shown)


Affie:
Nice. I would like to consolidate the tests a bit more now that this is in. We can merge an do as separate pr i think. 

Will other data entries work, do we want them to?Should the blob stores in dfg be persisted?

src/DataBlobs/services/BlobStores.jl
```diff
@@ -77,9 +77,13 @@ function addData!(dfg::AbstractDFG, blobstore::AbstractBlobStore, label::Symbol,
     return de=>db
 end
 
-function updateData!(dfg::AbstractDFG, blobstore::AbstractBlobStore, label::Symbol,  entry::AbstractDataEntry, blob::Vector{UInt8})
-    assertHash(entry, blob, hashfunction=hashfunction)
-    de = updateDataEntry!(dfg, label, entry)
+function updateData!(dfg::AbstractDFG, blobstore::AbstractBlobStore, label::Symbol,  entry::AbstractDataEntry, blob::Vector{UInt8}; hashfunction = sha256)
+    # Recalculate the hash - NOTE Assuming that this is going to be a BlobStoreEntry. TBD.
```
Affie, line 81: Yes. That should be save to assume. Entry type should become ::BlobStoreEntry

Affie: I’m not too sure about recalculating the hash though.

src/DataBlobs/services/BlobStores.jl
```diff
@@ -170,6 +174,25 @@ function addDataBlob!(store::FolderStore{T}, entry::BlobStoreEntry, data::T) whe
     end
 end
 
+function updateDataBlob!(store::FolderStore{T},  entry::BlobStoreEntry, data::T) where T
```
Affie, line 177: In other places update is warn then add if it exist.

GearsAD:
I think for now we can assume that a user would bind blobstores as needed. We can save them in, but then the question would be where should we bind them - on session, robot, or user? Let's see how that plays out.

src/DataBlobs/services/BlobStores.jl
GearsAD: Because you're updating the data itself, we need to recalculate the hash so that when you retrieve it next time, everything matches up. Or am I missing something? Will merge and we can discuss.

src/DataBlobs/services/BlobStores.jl
GearsAD: Good point, thanks!

GearsAD:
I think I'm going to force the CGDFG functions to only use BlobStoreEntry so that we don't have problems in the future with weird deserialization.

src/DataBlobs/services/BlobStores.jl
Affie: I would still go with `assertHash` the entry is not created in this function, but given by the user. So the assert is there to tell the user the blob does not match the entry.

Issue #582, opened 2020-08-09
loadDFG! error, "type does not have a definite number of fields", BlobStoreEntry


dehann:
[finalGraph.tar.gz](https://github.com/JuliaRobotics/DistributedFactorGraphs.jl/files/5047047/finalGraph.tar.gz)

On Julia 1.4.2

```julia
julia> using Flux
julia> using RoME
julia> dfg = initfg()
julia> loadDFG!(dfg, "finalGraph")  # or loadDFG!(dfg, "finalGraph.tar.gz")

sfolder = split(dstname, '.') = SubString{String}["/media/dehann/logs3/caesar/2020-08-09T05:30:25", "189/finalGraph", "tar", "gz"]
[ Info: loadDFG detected a gzip /media/dehann/logs3/caesar/2020-08-09T05:30:25.189/finalGraph.tar.gz -- unpacking via /tmp/caesar/random/6db72930-da24-11ea-229f-95665d6ce73b now...
ERROR: ArgumentError: type does not have a definite number of fields
Stacktrace:
 [1] fieldcount(::Any) at ./reflection.jl:705
 [2] #s16#65 at /home/dehann/.julia/packages/JSON2/ld4Kq/src/read.jl:443 [inlined]
 [3] #s16#65(::Any, ::Any, ::Any, ::Any, ::Any, ::Any) at ./none:0
 [4] (::Core.GeneratedFunctionStub)(::Any, ::Vararg{Any,N} where N) at ./boot.jl:526
 [5] read at /home/dehann/.julia/packages/JSON2/ld4Kq/src/read.jl:443 [inlined]
 [6] macro expansion at /home/dehann/.julia/packages/JSON2/ld4Kq/src/read.jl:259 [inlined]
 [7] read(::Base.GenericIOBuffer{Array{UInt8,1}}, ::Type{TimeZones.ZonedDateTime}; kwargs::Base.Iterators.Pairs{Union{},Union{},Tuple{},NamedTuple{(),Tuple{}}}) at /home/dehann/.julia/packages/JSON2/ld4Kq/src/read.jl:443
 [8] read at /home/dehann/.julia/packages/JSON2/ld4Kq/src/read.jl:443 [inlined]
 [9] macro expansion at /home/dehann/.julia/packages/JSON2/ld4Kq/src/read.jl:259 [inlined]
[...]

PR #588, opened 2020-08-11
Fix for non-millisecond timestamps

### LINKED ISSUE #560: CloudDFG timestamp bug
If no decimal for seconds.
ArgumentError: Unable to parse string "2020-07-27T00:27:34Z" using format dateformat"yyyy-mm-ddTHH:MM:SS.ssszzz". Unable to parse date time. Expected directive Delim(.) at char 20

GearsAD:
Fixes #560

### DIFF
--- a/src/services/Serialization.jl
+++ b/src/services/Serialization.jl
@@ -30,6 +30,9 @@ function unpackVariable(dfg::G,
         unpackBigData::Bool=true)::DFGVariable where G <: AbstractDFG
     @debug "Unpacking variable:\r\n$packedProps"
     label = Symbol(packedProps["label"])
+    # Make sure that the timestamp is correctly formatted with subseconds
+    packedProps["timestamp"] = replace(packedProps["timestamp"], r":(\d)(\d)(Z|z|\+|-)" => s":\1\2.000\3")
+    # Parse it
     timestamp = ZonedDateTime(packedProps["timestamp"])
     nstime = Nanosecond(get(packedProps, "nstime", 0))
     # Supporting string serialization using packVariable and CGDFG serialization (Vector{String})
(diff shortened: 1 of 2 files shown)


dehann:
Super cool!

src/services/Serialization.jl
```diff
@@ -30,6 +30,9 @@ function unpackVariable(dfg::G,
         unpackBigData::Bool=true)::DFGVariable where G <: AbstractDFG
     @debug "Unpacking variable:\r\n$packedProps"
     label = Symbol(packedProps["label"])
+    # Make sure that the timestamp is correctly formatted with subseconds
+    packedProps["timestamp"] = replace(packedProps["timestamp"], r":(\d)(\d)(Z|z|\+|-)" => s":\1\2.000\3")
```
Affie, line 34: Is the correct format not enforced by my change in ln 11:
`props["timestamp"] = Dates.format(v.timestamp, "yyyy-mm-ddTHH:MM:SS.ssszzz")`

As far as I could tell this was only a bug in CGDFG and worked correctly with forcing the format as above

dehann:
Don't think this method actually works, see here https://github.com/JuliaRobotics/DistributedFactorGraphs.jl/issues/582#issuecomment-671976222

src/services/Serialization.jl
dehann: 11?

dehann: Nope, see here: https://github.com/JuliaRobotics/DistributedFactorGraphs.jl/issues/582#issuecomment-671976222

Affie: Not the same thing

Affie: As far as I could tell it was working in v0.9, just not in the cloud serialization, which is separate.

PR #601, opened 2020-08-13
DFG version in serialized data

GearsAD:
Principal objects (variables and factors) should now have a version number embedded in it. For CGDFG we're embedding it in every object (because you may be loading sub-elements and you won't need to get the parent to figure out the version). This is currently used to emit a warning if older objects are loaded, but in the future we can use it to do automatic migrations.


Affie:
Maybe this helps: https://discourse.julialang.org/t/how-to-use-pkg-dependencies-instead-of-pkg-installed/36416

Affie:
Maybe:
`Pkg.dependencies()[Base.UUID("b5cc3c7e-6572-11e9-2517-99fb8daf2f04")].version`

src/DistributedFactorGraphs.jl
```diff
@@ -27,6 +27,7 @@ using JSON2 # JSON2 requires all properties to be in correct sequence, can't gua
 using LinearAlgebra
 using SparseArrays
 using UUIDs
+using Pkg
```
dehann, line 30: just a reminder to then also add `Pkg` to Project.toml and compat please

dehann: Pkg = "44cfe95a-1eb2-52ea-b672-e2afdf69b78f"

dehann: [compat]

Pkg = "1.4"

GearsAD:
Okay added a fallback for the Travis case - https://github.com/JuliaRobotics/DistributedFactorGraphs.jl/pull/601/files#diff-40d9a807e957cf49b7e74631b06ef686R17-R18 - it's passing now, will merge.

PR #607, opened 2020-08-13
export addVariable

dehann:


Affie:
I didn't export it on purpose. I saw it as experimental. Also, still not convinced about the parameters.

dehann:
thats fine thanks, we have to consolidate with Manifolds and must start somewhere. JuliaRobotics/RoME.jl#244 .  I've wanted to do macro's for a while but just not getting around to it.

PR #609, opened 2020-08-14
proper deprecation listSolvekeys,

### LINKED ISSUE #608: What happened to listSolvekeys?????
Deprecation warning or something? 
```julia
julia> listSolvekeys(fg)
ERROR: UndefVarError: listSolvekeys not defined
Stacktrace:
 [1] top-level scope at REPL[5]:1
```

dehann:
fix #608

### DIFF
--- a/src/Deprecated.jl
+++ b/src/Deprecated.jl
@@ -6,8 +6,6 @@
 ## Remove in 0.11
 ##==============================================================================
 
-export listSolvekeys
-
 @deprecate listSolvekeys(x...; kwargs...) listSolveKeys(x...;kwargs...)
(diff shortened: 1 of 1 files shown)


src/Deprecated.jl
```diff
@@ -5,6 +5,11 @@
 ##==============================================================================
 ## Remove in 0.11
 ##==============================================================================
+
+export listSolvekeys
```
Affie, line 9: FIY @deprecate automatically exports the name

dehann: oh, didnt know that thanks!

Issue #565, opened 2020-07-29
updateVariableSolverData! bug and API


Affie:
```julia
MethodError: no method matching updateVariableSolverData!(::LightDFG{SolverParams,DFGVariable,DFGFactor}, ::Symbol, ::Bool, ::Array{Symbol,1}, ::Bool)
```

and

```julia
MethodError: no method matching  updateVariableSolverData!(::LightDFG{SolverParams,DFGVariable,DFGFactor}, ::Symbol, ::Nothing, ::Symbol, ::Bool)
```

and this is also broken:

```julia
updateVariableSolverData!(dfg, xi.label, getSolverData(xi, :default), :graphinit, true, Symbol[], false)
```

Affie:
Add tests for this in DFG

Affie:
https://github.com/JuliaRobotics/DistributedFactorGraphs.jl/blob/21655bf6bcda16fe1c9b1f099f5c653a5b827ac5/src/services/DFGVariable.jl#L524

We should decide how we want to use this.

Affie:
originally posted in a comment: https://github.com/JuliaRobotics/DistributedFactorGraphs.jl/pull/570#discussion_r464301676

Maybe we shouldn't overwrite expected behaviour from base copy with deepcopy?
I expect `copy` to return a shallow copy and deepcopy for this.

Personally I don't think we should provide too many functions. Every function has to be documented and maintained.
The user can rather just do 
```juila
dcvnd = deepcopy(vnd)
dcvnd.selverKey = newKey
```

Is the function we are trying to replace `updateSolverData!(dfg, label, vnd, newsolvekey)`?
Where `vnd.solvekey != newsolvekey`.
Perhaps something like (although IIF needs to be updated slightly for it):
```julia
function deepcopySolverData!(dfg::AbstractDFG, label::Symbol, srcKey::Sybol, dstKey::Symbol)
    vnd = deepcopy(getSolverData!(dfg, label, srcKey))
    vnd.solverKey = dstKey
    updateSolverData!(dfg, vnd)
end  
```
Now that I'm thinking it through. `update` is maybe not the correct function to use. Its adding new solverdata so maybe:
```julia
function addSolverData!(dfg::AbstractDFG, label::Symbol, vnd, solverKey::Symbol)
    vnd.solverKey = dstKey
    addSolverData!(dfg, vnd)
end  
```

dehann:
[...]

PR #614, opened 2020-08-14
change updateVariableSolverData message

Affie:
see #565


dehann:
thanks for linking issue!

PR #616, opened 2020-08-14
check if pkg 1.5 fails IIF_TEST

dehann:


dehann:
looks like 1.5 is part of the problem. Can reopen this we get to it.

PR #635, opened 2020-08-19
SkeletonDFG constructor from CGDFG and use it to patch connectivity bug

Affie:


dehann:
Does this have anything to do with JuliaRobotics/Caesar.jl#279

Affie:
> Does this have anything to do with JuliaRobotics/Caesar.jl#279

I guess so, but is the check still needed now that iif can solve disconnected graphs?

PR #660, opened 2020-09-23
restore PackedType convert functions

### LINKED ISSUE #659: What API to get "PackedFactorType" from "FactorType"
Which function should be used downstream for this?  Closest I can find is the reverse in `services/Serialization.jl:decodePackedType`?  This used to just be a converter `convert(PackedInferenceType, Pose2Pose2) == PackedPose2Pose2`, looks like that might have been removed?

```julia
julia> convert(PackedInferenceType, LinearConditional(Normal()))
ERROR: MethodError: Cannot `convert` an object of type LinearConditional{1,Normal{Float64}} to an object of type PackedInferenceType
Closest candidates are:
  convert(::Type{S}, ::T) where {S, T<:CategoricalArrays.CategoricalValue} at /home/dehann/.julia/packages/CategoricalArrays/0ZAbp/src/value.jl:68
  convert(::Type{T}, ::T) where T at essentials.jl:171
Stacktrace:
 [1] top-level scope at REPL[202]:1
```

dehann:
closes #659

### DIFF
--- a/src/CloudGraphsDFG/services/CommonFunctions.jl
+++ b/src/CloudGraphsDFG/services/CommonFunctions.jl
@@ -179,7 +179,7 @@ function _structToNeo4jProps(inst::Union{User, Robot, Session, PVND, N, APPE, AB
             if fieldname == :solverData
                 fnctype = getSolverData(inst).fnc.usrfnc!
                 try
-                    packtype = convert(PackedInferenceType, fnctype)
+                    packtype = convertPackedType(fnctype)
                     packed = convert(PackedFunctionNodeData{packtype}, getSolverData(inst))
                     packedJson = JSON2.write(packed)
                     val = "\"$(replace(packedJson, "\"" => "\\\""))\"" # Escape slashes too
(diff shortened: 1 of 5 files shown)


src/CloudGraphsDFG/services/CommonFunctions.jl
```diff
@@ -133,6 +133,10 @@ Build a Cypher-compliant set of properies from a subnode type.
 Note: Individual values are serialized if they are not already.
 If the additional properties is provided, the additional
 properties will be added in verbatim when serializing.
+
+Related
+
+_Neo4jTo???
 """
 function _structToNeo4jProps(inst::Union{User, Robot, Session, PVND, N, APPE, ABDE},
                              addProps::Dict{String, String}=Dict{String, String}();
```
dehann, line 142: HI @GearsAD , could you perhaps update the partner `Neo4j-->Struct` function to also use the convert function as I have done here.  I was not able to find it in the code base :-P.  Whenever you get a chance and if it's relevant please.

PR #663, opened 2020-09-24
breaking change, AbstractRelative

dehann:


src/Deprecated.jl
```diff
@@ -125,3 +125,18 @@ function updateVariableSolverData!(dfg::AbstractDFG,
...
 end
+
+
+##==============================================================================
+## Deprecated in v0.11 Remove in the v0.12 cycle
+##==============================================================================
+
+export AbstractRelativeFactor, AbstractRelativeFactorMinimize
+
+@warn("BREAKING CHANGE v0.12: deprecating AbstractRelativeFactor, use AbstractRelativeRoots instead")
```
Affie, line 136: This looks like a good way to warn for deprecation. Perhaps it should be in __init so its called on using DFG?

dehann: good idea thanks, 9cf4494

PR #699, opened 2020-11-30
minimum GraphPlot = "0.4"

dehann:
Problem with compat, see JuliaRobotics/AprilTags.jl#57

This breaks Pkg compat quite badly, not sure where the issue is coming from.

just cc'ing @Affie on this too


Affie:
Hi, GraphPlot is an optional requirement with Requires.jl. The last time I checked there was no way to specify optional Requires.jl compact. see https://github.com/JuliaPackaging/Requires.jl/issues/82 and https://github.com/JuliaPackaging/Requires.jl/issues/3

Affie (changes requested):
GraphPlot is not in deps but a Requires.jl import, also see this comment: https://github.com/JuliaPackaging/Requires.jl/issues/3#issuecomment-529830411

Affie:
I previously thought of waiting for something like this (https://github.com/JuliaLang/Pkg.jl/issues/1285) to be implemented

dehann:
that's not the issue here.  i was trying to fix a compat issue by specifying a minimum, but not necessary.  this is a new non-resolvable set of pkg compat issue.

PR #700, opened 2020-12-01
softtype -> variableType

### LINKED ISSUE #603: Deprecation,  rename softtype -> VariableType
the idea of `softtype` has been removed, and is now just a regular hard type, so I think we should change the naming convention to `VariableType`.  Some of the API will need to be changed too, e.g. `getSofttype` will become `getVariableType` which is then similar to the existing `getFactorType`.

Also, the name `softtype` doesn't have any meaning really.  I added it a long time back for a different reason and use of the terminology `softtype` should probably expire.  The motivation was to allow different FactorsTypes to be added to the same VariableType, which increases reuse -- e.g. `MultivariateEuclid{2}` can be used on a `Point2, Pose2, Pose3, Point3`.  Becomes even more useful for quick tricks like `PartialEuclid{...}` (if you want to regularize one among variables).

Related:
- #577 
- #237

Affie:
close #603


dehann (commented):
Looks good thanks, just trying to figure out variableType vs VariableType quick.

dehann (approved):
looks good at variableType thanks

PR #702, opened 2020-12-07
Types are serialized into every CGDFG node, which is unnecessary and causing issues

GearsAD:
This resolves it by only serializing `_type` in for PPEs.

Setting this up for testing, need to figure out where everything is before merging.


GearsAD:
Wrong version... closing.

PR #704, opened 2020-12-10
Fixing tests for new IIF

GearsAD:
The downstream tests are borked - this should fix them.


test/iifCompareTests.jl
```diff
@@ -56,7 +56,7 @@ using Test
     # fg2 has been solved, so it should fail on the estimate dictionary
     @test !compareSimilarVariables(fg, fg2, skipsamples=true, skip=Symbol[:initialized;:inferdim])
 
-    tree = wipeBuildNewTree!(fg2)
+    tree = solveGraph!(fg2)
```
dehann, line 59: you could also just build tree with resetBuildTree!(fg)

dehann (approved):
one comment if you wanted, but looks good!

test/iifCompareTests.jl
GearsAD: yeah I couldn't find out what the new name was - thanks!

dehann: also the return from 'tree, smt, hist = solveTree!(fg)' is a Tuple

dehann: maybe we should consider a better name 'buildTreeReset!(fg)' which is more probable and better fits verbNouns

Affie (approved):

PR #705, opened 2020-12-10
Fixing an issue with the _type field in CGDFG

GearsAD:
Merge after #704


test/iifInterfaceTests.jl
```diff
@@ -141,9 +141,9 @@ end
     #FIXME don't know what it is supposed to do
     @test_broken lsfTypes(dfg)
 
-    @test ls(dfg, LinearConditional) == [:abf1]
-    @test lsf(dfg, LinearConditional) == [:abf1]
-    @test lsfWho(dfg, :LinearConditional) == [:abf1]
+    @test ls(dfg, LinearRelative) == [:abf1]
+    @test lsf(dfg, LinearRelative) == [:abf1]
+    @test lsfWho(dfg, :LinearRelative) == [:abf1]
```
dehann, line 146: new name is:
'lsfTypes(fg, LinearRelative)'

dehann (commented):
just the lsfTypes thing

test/iifInterfaceTests.jl
Affie: @dehann There is no function like that.
Only lsfTypes(dfg) -> Return `Vector{Symbol}` of all unique factor types in factor graph.

Did you perhaps mean just  `lsf` as the test just above that?

Should we deprecate `lsfWho` in favour of just `lsf`

PR #718, opened 2020-12-16
Fixing listVariable and listFactor with regex

GearsAD:
Not the most elegant solution, may be other places that need fixing, but I need this is master as soon as possible. Please let me know if there are similar issues causing any trouble.


dehann (approved):
what could possibly go wrong

Issue #728, opened 2021-01-18
New error in tests, todot


dehann:
During regular testing:

```julia
Producing Dot Files: Test Failed at /home/dehann/.julia/dev/DistributedFactorGraphs/test/testBlocks.jl:1414
  Expression: todota || todotb
Stacktrace:
 [1] ProducingDotFiles(::Type{T} where T, ::Nothing, ::Nothing, ::Nothing; VARTYPE::Type{T} where T, FACTYPE::Type{T} where T) at /home/dehann/.julia/dev/DistributedFactorGraphs/test/testBlocks.jl:1414
 [2] ProducingDotFiles at /home/dehann/.julia/dev/DistributedFactorGraphs/test/testBlocks.jl:1394 [inlined] (repeats 2 times)
 [3] top-level scope at /home/dehann/.julia/dev/DistributedFactorGraphs/test/interfaceTests.jl:178
 [4] top-level scope at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.5/Test/src/Test.jl:1115
 [5] top-level scope at /home/dehann/.julia/dev/DistributedFactorGraphs/test/interfaceTests.jl:176
```

Issue #725, opened 2021-01-15
julia v1.6-beta1 bug


Affie:
xref https://github.com/JuliaRobotics/IncrementalInference.jl/issues/1120

dehann:
This is fine to close right?

PR #730, opened 2021-01-20
Converted travis to github ci

jim-hill-r:
Fixes #729

Still needs
- [ ] Testing in github to validate it works
- [x] Updated badging

PR #731, opened 2021-01-20
Bugfixes for julia 1.6.0-beta1

Affie:


dehann:
HI @Affie , just in case, dont wait on #730 -- if tests pass locally by you then merge this so work can continue on fixing JuliaRobotics/IncrementalInference.jl#1120

Affie:
Ok, I’ll merge so long. It’s passing more than current master, but not everthing. I would have liked all to pass, but travis is too much of a pain.

dehann:
also closes #728

dehann:
also closes #725

PR #741, opened 2021-02-14
deprecate old code

dehann:


dehann:
::Vector{P} upgrade for DFG v0.15 is more important at this time, so these deprecations will be deferred for another cycle.

dehann:
will redo in alternative PR

PR #768, opened 2021-05-21
compat AbstractManifold

dehann:


dehann:
to fix JuliaRegistries/General#37200

dehann:
Manifolds.jl must first support compat for ManifoldsBase.jl v0.11

Issue #763, opened 2021-05-17
VariableNodeData discussion for upgrading to Manifolds.jl


Affie:
To start a discussion on how best to deprecate the `val` and `bw` types of `Array{Float64,2}` to a vector of type `P` where `P` is the type of a point on a manifold. Or possibly something different? 

Serialization should also be considered.

A possible way is to use a `Union` as a temporary transition (although it might be slower) and then change to a type
```julia
mutable struct VariableNodeData{T<:InferenceVariable}
    """#TODO will become Vector of a point on the manifold, 
          but should be backward compatible with Array{Float64,2}"""
    val::Union{<:AbstractVector, Matrix{Float64}}
    """#TODO will become Vector of a bandwidth/covariance of a point, 
          but should be backward compatible with Array{Float64,2}"""
    bw::Union{<:AbstractVector, Matrix{Float64}}
    ...
end
```
And then it can become:
```julia
mutable struct VariableNodeData{T<:InferenceVariable, P, B}
    "Vector of a point on the manifold"
    val::Vector{P}
    "Vector of a bandwith/covariance of a point"
    bw::Vector{B}
    ...
end
```
We can also 2 more fields and then just swop over from the one to the other.

I looked at TensorCast, but the Manifold points do not fit in with the `Array{Float64,2}` e.g. SO(2) is a Vector of 2x2 matrices.

dehann:
xref JuliaRobotics/IncrementalInference.jl#1242

> P is the type of a point on a manifold. Or possibly something different?

I agree, storing some `<:Vector` of points `p::P` each being of `<:ManifoldBase`. 

I think the Union approach is probably easiest -- that way we will hopefully get all the errors for any lingering usage of `.val`.  we should probably do the warn once on getfield think again to be sure.

I'm less confident about what is going to happen to the `.bw` field as we progress -- going with `::Vector` should be good enough to make the transition... 

> TensorCast

that comment was just for convenience, the packages do not have much to do with each other at this point in time.  (tensors and manifold operations are friends though...)

mateuszbaran:
Instead of `Vector` of points you may also consider using our power manifold: https://juliamanifolds.github.io/Manifolds.jl/stable/manifolds/power.html . It has the benefit of directly supporting, for example, a vector of points on SO(2) as a `2x2xN` matrix, so you can use that with TensorCast, LoopVectorization or other generic libraries.

kellertuer:
[...]

Issue #590, opened 2020-08-11
Serialization and Marshaling (again)


Affie:
Just capturing ideas for discussion.

## History
> Another thing, we actively decided after a 2 year process not to define the "schema" for serialization. We originally had a protobuf packing approach which was hard schema type dependent and which supported forward-backward multilanguage compatibility. After much pain, we decided that top level packing MUST be JSON, and that the JSON string must at top level include the format in which the data is serialized. This is to mimic how the Internet has standardized it's data exchange format. So StructTypes.jl might be a real problem given the history on serialization.
So to stress, its probably not a good idea going down a path to unilaterally "change everything" on serialization.
In this case, it will be better to understand what the problems are with JSON.jl JSON2.jl JSON3.jl, because clearly they haven't figured it all out either. But we likely won't go down writing JSON4.jl either.
Also remember that the whole point of JSON is that serialization should be marginally or completely available outside Julia. The ZMQ interface for example is about multilanguage support and it uses the same serialization. The DB storage is definitely not a Julia implementation, later we will add a FuseDFG driver, etc, etc. A little while back we came up with the requirement that ZMQ messages should be exactly the same as FileDFG content. The schema discussion becomes a whole issue there. There are 100 different schema's available... So serialization only has 2nd best answers for everything. Whenever you pick the "best" answer for one requirement, then all others get real upset. So we have to accept a 2nd best solution for everybody and JSON is the best bet at this time.

**So JSON**

### JSON.jl + unmarshal.jl
...

### [JSON2.jl ](https://github.com/quinnj/JSON2.jl)
Deprecated in favor of JSON3

### [JSON3.jl](https://quinnj.github.io/JSON3.jl/stable/)
[See this](https://quinnj.github.io/JSON3.jl/stable/#DataTypes-1) for the 2 trait options: `Struct()` or `Mutable()`
In short Struct() is fast, but the order is important (So has to be preserved if created/modfied from outside).

Example:
```julia
julia> using JSON3
julia> using DistributedFactorGraphs
julia> using TimeZones
julia> using UUIDs
julia> bse = BlobStoreEntry(:a, uuid4(), :abc, "", "", "", "", ZonedDateTime("2020-08-11T14:32:27.000+02:00"))
[...]

PR #783, opened 2021-06-19
towards val::Vector{T}

dehann:
fyi @Affie , im changing @defVariable slightly to take the identity point for a variable rather than just a point type.  We still get the same point type information using `typeof`.  Let me know if there is a problem.  

Issue is we need for serialization something like `getCoordinates` and `getPoint`.  For the default dispatch, im just pointing through to the basic Manifolds `get_coordinates` and `get_vector` functions.  Users doing something more fancy should override DFGs `getCoordinates` and `getPoint` functions if needed.


Affie:
Hi, I also considered it, but as far as i know the identity element is not general on all manifolds.

dehann:
Okay, but we need something for serialization.  I think getCoordinates is more general than requiring users dedicated serialization converters for each manifold?  Hence the need for an easy Identity element.

Besides, I don't think it is that big a difference between writting:
```julia
@defVariable SpecialEuclidean(3) ProductRepr([0;0;0.0], [1 0 0; 0 1 0; 0 0 1.0])
```
and

```julia
@defVariable SpecialEuclidean(3) ProductRepr{Tuple{Vector{Float64}, Matrix{Float64}}}
```

If you really don't like it, can we at least just use if as a transition mechanism, until we find something better than `DFG.getCoordinates / Manifolds.get_coordinates`?

dehann:
> as far as i know the identity element is not general on all manifolds

probably not, and in these cases we can just require users to define their own `DFG.getCoordinates(::Type{<:InferenceVariable, p)` and `DFG.getPoint(::Type{<:InferenceVariable}, coords)`.  As default, all manifolds that are also a group can just go on defaults -- i.e. groups always have an identity elements although general manifolds might not have.

dehann:
I'm going to continue down this path for now, to get IIF and RoME to also work for `vnd.val::Vector{P}`.  I'm thinkging that the last disruption will be to maintain `IIF.getPoints`.  There are a lot of places in the downstream code where I used:
```julia
matrixofcoords = getBelief(fg, :x1) |> getPoints
```

So I was thinking to use the same `DFG.getCoordinates` default as transition method for this `getPoints` function.  **Note** the difference between `getPoint`above and `getPoints` here.

dehann:
Oh and forgot to mention the tests here DFG should be passing on this PR at current commit 6e253d0

Affie:
Just to take a step back, why is the "identity element" better for serialization?
Also, why would you need coordinates?

dehann:
How do we store something like `ProductRepr([1;2], [0 1; 0 1]` in FileDFG or CloudDFG?

Currently all our plumbing is setup for a long vector of Float64 for serialized.

src/entities/DFGVariable.jl
```diff
@@ -16,58 +16,92 @@ Data container for solver-specific data.
...
 $(TYPEDFIELDS)
 """
-mutable struct VariableNodeData{T<:InferenceVariable}
-    val::Array{Float64,2}
-    bw::Array{Float64,2}
-    BayesNetOutVertIDs::Array{Symbol,1}
-    dimIDs::Array{Int,1} # Likely deprecate
+mutable struct VariableNodeData{T<:InferenceVariable, P}
+    val::Vector{P}
+    bw::Vector{Vector{Float64}}
```
Affie, line 21: Why this structure? I also use it as a Matrix for the covariance in parametric.

dehann:
PS, the majority of work for this is upgrade is in IIF, so i'm happy to do it in two iterations, but would just like to keep moving rather than grinding away in DFG too long

dehann:
see here:
https://github.com/JuliaRobotics/DistributedFactorGraphs.jl/pull/783/files#diff-78949ed1d5af6333167943ed87149739639a1ef72095cc96b0746a15f8024f04R217-R223

src/services/DFGVariable.jl
```diff
@@ -129,6 +131,48 @@ Interface function to return the manifold point type of an InferenceVariable, ex
...
+Related
+
+[`getPoint`](@ref)
+"""
+function getCoordinates(::Type{T}, p) where {T <: InferenceVariable}
+    M = getManifold(T)
+    p0 = getPointIdentity(T)
+    X = log(M, p0, p)
+    get_coordinates(M, p0, X, DefaultOrthonormalBasis())
+end
```
Affie, line 173: I'm not sure about this. It's not general and makes a lot of assumptions. 

Perhaps it can be more accurate with a new abstract `LieGroupInferenceVariable <: InferenceVariable`  (but also don't like it that much)

src/entities/DFGVariable.jl
dehann: oh i just duplicated val to bw.  We can leave bw as Matrix for now.  Sorry I'll change that back

src/services/DFGVariable.jl
dehann: Lets not pre-optimize.  I'm trying to take a reasonable step towards Manifolds based on where we are today.  `Manifolds.get_coordinates` already exists, so my feeling is we can leverage that as default and user must override for general cases.

For this step, I'm totally against building new abstracts.  Maybe later once `vnd.val::Vector{P}` is already working fully downstream.

Affie:
> How do we store something like ProductRepr([1;2], [0 1; 0 1]) in FileDFG or CloudDFG?
Currently all our plumbing is setup for a long vector of Float64 for serialized.

Maybe I'm missing something, is it not as simple as:
```julia
julia> A = SA[1 2; 3 4]
2×2 SMatrix{2, 2, Int64, 4} with indices SOneTo(2)×SOneTo(2):
 1  2
 3  4

julia> v = vec(A)
4-element SVector{4, Int64} with indices SOneTo(4):
 1
 3
 2
 4

julia> B = typeof(A)(v)
2×2 SMatrix{2, 2, Int64, 4} with indices SOneTo(2)×SOneTo(2):
 1  2
 3  4
```

dehann:
Not everything is `ProductRepr`, so each variable type will have its own unique way of going between "coordinates" that we can stack in a vector, and the "group" that we use for computations in VND.

We can change all this after the first iteration `vnd.val::Vector{P}` is working end to end.  I'm sure there are better ways.  I'm just taking the most direct route to getting the upgrade with least breakages in IIF, RoME, and Plotting

dehann:
busy changing bw back to matrix.. will commit in a few minutes

Affie:
> We can change all this after the first iteration vnd.val::Vector{P} is working end to end. I'm sure there are better ways. I'm just taking the most direct route to getting the upgrade with least breakages in IIF, RoME, and Plotting

I'm all for getting the ball rowling. As long as it doesn't get baked in because of serialization. I would rather design how it would work best for numerics and then have serialization fit in with that. 

So I say go ahead, as you mentioned the hard part will be downstream and we need to get to it.

Affie:
If we can serialize:
- ProductRepr (which is just a tuple) 
- Vector
- Matrix
- S/MVector
- S/MMatrix

Is that not enough? If a user wants some other type they can implement a new serialize function

dehann:
> rather design how it would work best for numerics and then have serialization fit in with that. 

I agree with that, we will revisit #590 so we can be a bit more loose with serialization now.  I want to include in this design criteria how easy it is for a newcomer to build new variables and factors.

> So I say go ahead, as you mentioned the hard part will be downstream and we need to get to it.

Cool, lets tag downstream working, and then come back to DFG and do a deprecation cycle for cleanup on the Manifolds.jl upgrade.

> I took matrix out

Think we (our early users) are very likely to want matrices, so should include that.

> If we can serialize:
> Is that not enough? If a user wants some other type they can implement a new serialize function

Yup, 100% agree.  One thing I've wanted to do for a long time is make variables and factors use a `@AutoSerializeDFG` macro and avoid the need for users to have to constantly specify:
- PackedDistribution
- PackedVariable
- PackedVariableNodeData
- PackedFunctionNodeData
- PackedManifoldPoint
- etc.

I'd say we do that as a cleanup, keeping bits of JSON3, #590, etc in mind;  but that is after this work to transition to Manifolds.jl proper, and also get JuliaRobotics/IncrementalInference.jl#1010 resolved first.

Affie:
xref #763

Affie:
Also, have a look at https://juliamanifolds.github.io/Manifolds.jl/dev/manifolds/power.html#NestedReplacingPowerRepresentation

dehann:
xref JuliaManifolds/Manopt.jl#90

PR #788, opened 2021-07-06
Update MeanMaxPPE to a point on manifold

Affie:
TODO serialization of MeanMaxPPE

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "DistributedFactorGraphs"
 uuid = "b5cc3c7e-6572-11e9-2517-99fb8daf2f04"
-version = "0.15.0"
+version = "0.15.1"
 
 [deps]
 Base64 = "2a0f44e3-6c83-55bd-87e4-b1978d98bd5f"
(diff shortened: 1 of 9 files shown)


Affie:
What is the reason to add the variable type to ppe? won't it always be attached to a variable that has the type?

Also: why is the type included as a field and a type parameter? Is a type parameter not enough. Or a field only, but I don't know if `::Type{T}` is type stable.

dehann:
> What is the reason to add the variable type to ppe? won't it always be attached to a variable that has the type?

Previous conversation where we decided to duplicate on purpose, so that when a user gets a PPE in non-Julia world, they can back out what the values mean.  PPE and VND might not always travel together.

> Also: why is the type included as a field and a type parameter? Is a type parameter not enough. Or a field only, 

Only should be good enough, lets perhaps track that in an issue and figure that decision out when the manifolds and IIF 1010 work simmers down a bit?

> but I don't know if ::Type is bad for performance.

I also don't know.  The first thing that comes to mind for me is if abstract types and type inference is needed when working with types.  The second thing that comes to mind is that Base already uses the idea of Type for converters, eg convert(;:Type, obj) so it cannot be that bad for performance?

dehann:
This is sort of coming up, adding xref here for future design:
- https://github.com/JuliaRobotics/IncrementalInference.jl/issues/1507

Issue #527, opened 2020-07-06
Remove all return types?


dehann:
As before, we are removing return types, correct?

https://github.com/JuliaRobotics/DistributedFactorGraphs.jl/blob/master/src/services/AbstractDFG.jl

dehann:
From Slack:

dehann
Johan and i voted, 2-0 for no return types :-). the reason is that return types in Julia work different from C.  The most troubling is when you convert a healthy concrete type up to an abstract type, which is happening a lot in the code at the moment, then you loose all the static operation and replace good code with dynamic dispatch around the forced abstract types.  Furthermore, forcing a return type inserts a convert function in all cases which ends up producing really difficult error messages on a failure.  So using return types doesn’t buy you anything in terms of type stability, it actually takes away.  The way to identify type issues is with @warn_type

@Affie 
I’ve been taking out the abstract return types, but not the rest. Are we removing all of them? Ideally the convert should be optimized out and asset can be useful, but it does lead to difficult error messages. The c programmer in me still likes it and there are places that it helps. I agree it helps with understanding the code. So maybe we should move it to the doc strings where it’s not needed.
I don’t think @warn_type will do anything. Calling assert on the return type is a bit closer.
We have also been changing arrays to abstract arrays and floats64 to real in a lot of places to allow more flexibility. It was needed for automatic differentiation in my case.

dehann:
> So maybe we should move it to the doc strings where it’s not needed.

Agree, at minimum lets do docstrings.

> Calling assert on the return type is a bit closer.

That would be for the developer only, and likely reduce performance a bit, so maybe a comment line with the assert in (at the end of a function) so when developing it can be uncommented if needed?

> It was needed for automatic differentiation in my case.

oh thats news to me, by "it" do you mean the hard return type was needed?

For Julia in general, a function should arrive at a type stable solution (which is where `@warn_type` helps).  Forcing a return type does not help the type inference as far I'm aware.

Affie:
> oh thats news to me, by "it" do you mean the hard return type was needed?

no, the opposite. It had to be abstract input types to work.


> That would be for the developer only, and likely reduce performance a bit, so maybe a comment line with the assert in (at the end of a function) so when developing it can be uncommented if needed?

@assert is a bit more complicated, from the docs:
[...]

PR #801, opened 2021-08-23
drop bad (forced) function level return types

dehann:
Hi, @Affie  @GearsAD , just a heads up -- I'm busy cleaning up code as I see things.  On forced return types, its better to force the return statement itself if needed (not the whole function).  It's especially bad if the whole function return is forced to a non-concrete type.


Affie:
Jip the decision was made a while ago to remove them. These must have just been missed. Xref #527

PR #803, opened 2021-09-16
fixing a show (should be display)

dehann:


GearsAD:
...some dumps are too long...

Affie:
This is by design: "MIME type = human readable, 2-arg = parseable, compact = keep it short"

dehann:
@GearsAD , Johan says we should use `display(::LightDFG)` in the duplicator `show` (not `show(::LightDFG)`).  Johan understands why, based on the patterns he learned from `Base`.  I'm sure Johan is right.

PR #806, opened 2021-10-15
enh api on defVariable and typeModName

dehann:


src/services/DFGVariable.jl
```diff
@@ -102,6 +102,8 @@ macro defVariable(structname, manifold, point_identity)
 
         DFG.getPointIdentity(::Type{$structname}) = $point_identity
 
+        DFG.getVariableType(::typeof($manifold)) = $structname
```
Affie, line 105: @dehann,  this is a bug, what was the intention here?

Issue #809, opened 2021-10-17
CGDFG infinite recursion on deleteFactor


GearsAD:
The IIF current solver behavior replaces factors during `rebuildFactorMetadata`. This is an issue with CGDFG because deleting factors retrieves the factor. To fix this we should introduce a flag to suppress the `getFactor` call in `deleteFactor`, which is only used by IIF during `rebuildFactorMetadata`.

dehann:
xref JuliaRobotics/IncrementalInference.jl#1424

dehann:
think we can close this and resolve the DFGVariable[] IIIF 1424 issue over there?

GearsAD:
Yep, merging

PR #810, opened 2021-10-17
Adding flag to deleteFactor

GearsAD:
Addresses #809. Need to release this before applying the change to IIF.


dehann:
xref JuliaRobotics/IncrementalInference.jl#1424

PR #812, opened 2021-10-17
don't force return types, esp non-concrete type

dehann:


GearsAD (approved):

GearsAD:
Should also go in the alias below that - https://github.com/JuliaRobotics/DistributedFactorGraphs.jl/pull/812/files#diff-9bf1902b28a177f5f05513a0cd3f168a0954a08a3a463ed8c7d5413deb4b29d5R382

Issue #818, opened 2021-11-13
issue with GraphPlot, spring_layout


dehann:
Assuming this is some upstream change

```julia
Plotting Tests: Error During Test at /home/runner/work/DistributedFactorGraphs.jl/DistributedFactorGraphs.jl/test/runtests.jl:60
  Got exception outside of a @test
  LoadError: MethodError: no method matching spring_layout(::FactorGraph{Int64, DFGVariable, DFGFactor})
  Closest candidates are:
    spring_layout(::Graphs.AbstractGraph) at /home/runner/.julia/packages/GraphPlot/pvheh/src/layout.jl:104
    spring_layout(::Graphs.AbstractGraph, ::Integer, ::Any...) at /home/runner/.julia/packages/GraphPlot/pvheh/src/layout.jl:177
    spring_layout(::Graphs.AbstractGraph, ::Any) at /home/runner/.julia/packages/GraphPlot/pvheh/src/layout.jl:104
    ...
  Stacktrace:
    [1] gplot(g::FactorGraph{Int64, DFGVariable, DFGFactor}; layout::typeof(spring_layout), keyargs::Base.Iterators.Pairs{Symbol, Vector{T} where T, Tuple{Symbol, Symbol, Symbol}, NamedTuple{(:nodelabel, :nodesize, :nodefillc), Tuple{Vector{String}, Vector{Float64}, Vector{ColorTypes.RGB{FixedPointNumbers.N0f8}}}}})
      @ GraphPlot ~/.julia/packages/GraphPlot/pvheh/src/plot.jl:229
    [2] dfgplot(dfg::LightDFG{NoSolverParams, DFGVariable, DFGFactor}, p::DFGPlotProps)
      @ DistributedFactorGraphs.DFGPlots ~/work/DistributedFactorGraphs.jl/DistributedFactorGraphs.jl/src/DFGPlots/DFGPlots.jl:65
    [3] dfgplot(dfg::LightDFG{NoSolverParams, DFGVariable, DFGFactor})
      @ DistributedFactorGraphs.DFGPlots ~/work/DistributedFactorGraphs.jl/DistributedFactorGraphs.jl/src/DFGPlots/DFGPlots.jl:53
    [4] top-level scope
      @ ~/work/DistributedFactorGraphs.jl/DistributedFactorGraphs.jl/test/plottingTest.jl:18
    [5] include(fname::String)
      @ Base.MainInclude ./client.jl:444
[...]

PR #819, opened 2021-11-15
bump v0.17.0

dehann:
The following commits were accidentally pushed to main branch by mistake, so testing here:
- 1ab04d3
- 39a1dd4

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "DistributedFactorGraphs"
 uuid = "b5cc3c7e-6572-11e9-2517-99fb8daf2f04"
-version = "0.16.2"
+version = "0.17.0"
 
 [deps]
 Base64 = "2a0f44e3-6c83-55bd-87e4-b1978d98bd5f"


dehann:
Hi @Affie , i tried to fix new dfgplot issue #818 but had some difficulty with the test.  I'm skipping the test here 39a1dd4 and keeping focus on marine example.  If you have spare time at some point, could you perhaps take a look please -- I'm hoping you can figure out whats up with new GraphPlot faster. 

PS, I'd suggest we make the test files more "standalone" so that we can easily run on their own to help others debug more quickly.  I tried the test file, but then `TestFunctorInferenceType1` was not define, so then I found and defined that, but then a next error was that `.usrfnc` was not defined.  I know `usrfnc` is somewhere in CCW, but that is only defined downstream in IIF, this was before even getting to the #818 so I decided to just skip the plotting issue for another day.  This is the line that is failing before getting to 818
https://github.com/JuliaRobotics/DistributedFactorGraphs.jl/blob/ccd60c4d94fada0cdfc748cad22264aab9115809/test/plottingTest.jl#L17

PR #875, opened 2022-04-12
Defensively implementing backticks on sentinel nodes (userId, robotId, sessionId)

### LINKED ISSUE #870: UserID must be backticked in Neo4jDFG
This allows userIDs to be free strings.

GearsAD:
Resolves #870

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "DistributedFactorGraphs"
 uuid = "b5cc3c7e-6572-11e9-2517-99fb8daf2f04"
-version = "0.18.2"
+version = "0.18.3"
 
 [deps]
 Base64 = "2a0f44e3-6c83-55bd-87e4-b1978d98bd5f"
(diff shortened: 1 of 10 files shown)


Affie (approved):
I just read through it, but LGTM

Affie:
We should just bump the minor version.

Affie:
Is it not a breaking change, so v0.19?

jim-hill-r:
I'll do some tests this evening. My only immediate thought is what is actually stored in the db. The db doesn't store backticks right, it's just syntax for the query? That means it's non breaking yes?

Affie:
I'll leave that up to you, happy either way.

GearsAD:
This isn't a breaking change and it doesn't change how the data is persisted.

dehann (approved):
Thanks doing this Sam!

dehann:
If not breaking then we can just do a patch release v0.18.3.

jim-hill-r (approved):

jim-hill-r:
Testing looked great! I was able to run a hex and a solve with no problems using user "guest@navability.io". I then did a hex and solve with "jim@navability.io" and that also went well. I also did a bunch of AprilTag things via Introduction then checked the viz. Both fg and geometric work. Go forth and release the patch.

GearsAD:
Great, merging and tagging!

PR #885, opened 2022-07-12
User OrderedDict to store variables and factors

Affie:
@dehann, what do you think about using `OrderedDict`s? I don't think this change will affect anything downstream, but still have to check.


Affie:
This has the added benefit of a better user experience, for example:

```jullia
julia > fg = generateGraph_Hexagonal();

julia> ls(fg)
8-element Vector{Symbol}:
 :x0
 :x1
 :x2
 :x3
 :x4
 :x5
 :x6
 :l1

dehann:
Side note, there already is a variable add history (maybe factors too), which stores similar information.:
https://github.com/JuliaRobotics/DistributedFactorGraphs.jl/blob/1b44e72ece6db63f8fb33f2a6a528318ee7c4ba6/src/services/AbstractDFG.jl#L59

Either way, i'm good to change to `OrderedDict`.  Might make performance a little slower on adding to a large graph, but benefits outweigh that need.

PS, parametric tree solve will have smaller subgraphs of unordered graphs, so any sequencing of factors for a power manifold etc. will need to be done more carefully.  Luckily those graphs should be much smaller, so hearty sorting of factors should be fine in that context.

dehann:
> only an ordered dictionary, not a sorted one.

Oh right, you're correct

dehann:
> Perhaps we can merge to master and see if the dev branches tests pass.

sure, im good to merge.

PR #894, opened 2022-08-22
Point type parameter in DFGVariable and vnd.covar

### LINKED ISSUE #884: Add covariance in VND container for parametric and move away from using `.bw` for both
It might be easier to have a dedicated place for this. It can also be used to track multiple parametric modes.

### LINKED ISSUE #979: variableType<:InferenceVariable  in VND
Going further, why did we have `VND{T <: InferenceVariable}`, and not just `DFGVariable{T}`?  This is more of a DFG consideration.

_Originally posted by @dehann in https://github.com/JuliaRobotics/IncrementalInference.jl/issues/1228#issuecomment-826172622_

Affie:
close #979
close #884


dehann (approved):

Affie:
This looks to have slowed down parametric solve a bit

src/services/Serialization.jl
```diff
@@ -121,25 +121,25 @@ function unpackVariableNodeData(d::PackedVariableNodeData)
...
-        d.dontmargin, 
-        d.solveInProgress, 
-        d.solvedCount, 
-        Symbol(d.solveKey),
-        Dict{Symbol,Threads.Condition}() )
+    return VariableNodeData{T, getPointType(T), getDimension(T)}(;
+        id = d.id,
+        val = vals, 
+        bw = BW, 
+        # covar = #FIXME serialize covar
```
Affie, line 128: Note, `covar` is not serialized yet

Affie:
For packing:
TODO from https://github.com/JuliaRobotics/DistributedFactorGraphs.jl/pull/981:
- [ ] This should be Union{Vector{Float64}, Nothing}
- [ ] When going from packed -> full this should convert nothing -> []
- [ ] When going from full -> packed no extra changes needed
- [ ] Functional code need to check for []

Affie:
@GearsAD, I'm not sure about what to do on your TODO, as long as a missing array is handled as [] in json it should work.
Will the field be missing from json?

PR #906, opened 2022-10-31
add mergeDataEntries!

dehann:


Affie:
I also think merge is the correct thing to do here as apposed to copy. 
Just 2 things:  
- I think merge should follow julia: "If the same key is present in another collection, the value for that key will be the value it has in the last collection listed"
- Please add tests

PR #936, opened 2023-01-12
Remove LightDFG and deprecate with error

Affie:
@GearsAD, I didn't test since I don't know if you already did this, but let me know if I should carry on.


Affie:
I missed LightGraphs

Affie:
It looks like something unrelated with serialization is broken.

PR #937, opened 2023-01-13
add transcodeType util for super unmarshal

dehann:
Add transcodeType as a sort of super unmarshal function.  Keep coming up against the need for something like this, so just adding it with tests.

Superseeds previous upstream attemps:
- https://github.com/lwabeke/Unmarshal.jl/issues/39#issuecomment-1073183609


dehann:
merged in since this only adds, should have very little conflict, if any.

PR #942, opened 2023-01-29
Standardizing DFG objects against new patterns

GearsAD:
Updates for new drivers:
- [X] Use label not ID for User/Robot/Session
- [X] Clean up blob storage

WIP here, can merge once this is done.


GearsAD:
@dehann DataEntry updates are here

dehann:
Oh cool!  

On the JSON3 update, if that is going to break compat with DFG 18 serialized content, can we bump minor version here to v0.20 instead (since that would be a breaking change).   Nevermind if JSON3 is not going to break compat, then v0.19.x is good.

dehann (approved):

PR #953, opened 2023-02-02
copyGraph recontr on destDFG instead

dehann:


Affie (changes requested):
Neighbors should still be fetched from the source graph.

Issue #904, opened 2022-10-11
get nodes call should support getnodekwargs


dehann:
All getVariable and getFactor calls should have a `; getnodekwargs...` parameter to allow more advanced downstream use.

PR #964, opened 2023-02-28
kw on getVar getFct,

dehann:
close 
- #904


Affie:
Would this not allow errors to go unnoticed? If you think you have the correct kwarg name and made a spelling mistake it won't be picked up.

dehann:
oaky, lets go without it.  I had use cases downstream that needed something like this.  Can just add in a patch release later when it comes up.

Affie:
Ok, I would be curious as to when this will be needed.

PR #991, opened 2023-03-23
Changes needed for NvaSDK v0.6

Affie:


Affie:
I'm not following. There is only one definition. If you are talking about the propertynames overwrite, it's just to make other stuff easier. but I can take it out and make a new function that does the same thing, if you don't like it?

dehann:
yes, the property names.  Perhaps you have a clear case on all user seeing all properties is making it harder?

> just to make other stuff easier. but I can take it out and make a new function

Im against that: I really think we should stay away from anything that might indicate `BlobEntry` has (A set of fields) here or (B set of fields) there.  Just one struct with its long list of fields system wide.  Calls like `getfieldnames` or `getproperty` should not hide stuff in anywhere.  My motivation is over the past 2 years around 6 flavors of `BlobEntry` popped up, and the work now is to drop that all just one everywhere and avoid the temptation of "make a quick copy of just the stuff I need" and then building new code against a `less-than-BlobEntry`.  Any extraneous fields should just be optional with defaults.

Affie:
I deleted all propertynames overwrites form this pr

Affie:
I'm assuming its ok to merge now. so gooing ahead

PR #995, opened 2023-03-24
Change meaning of nstime

Affie:
I've been using it like this, any objections if we change the definition? Your epoch can be the last timestamp second making it as it was.


GearsAD:
Up to you guys, I'll make it work with large numbers either way. Note it's on both variables and factors.

Affie:
> Note it's on both variables and factors.

jip, and I would like it on `BlobEntries` as well. but will use metadata for now.

GearsAD:
On BlobEntry, can do - fire up an issue on cloud-system. I’d like to knock
them out while the engine is open and on the floor.

dehann:
I've been using nanosecond in the same ROS does -- i.e. nanosecond is always smaller than 1_000_000_000 and only accounts for the high fidelity timing information.  That allows the seconds counter to work with unix time and not loose resolution in Float64 usage.  So you can always get the accurate time estimate by
```
ns_accurate_time = obj.secs + 1e-9*obj.nsecs
```

Does that work with the change you suggesting @Affie, since I cannot really make out what you mean in the new docstring?  What I understand from the suggested docstring is (but this doesnt seem right):
```
User A is using epocs say every 2.5 seconds for some situation, therefore nstime is 2_500_000_000, since user B would have no way of knowing what that for?
```

dehann:
We'll probably have to talk about this more on call.  Not sure the text comms is working here.

dehann:
what i mean is that there should be no duplication in meaning or value between what is stored in secs vs nsecs.  And secs already stores the time from start of trajectory, so nsec should not be a redundant register for time from start of trajectory or time from user keyframe epoch etc.

PR #996, opened 2023-03-27
new getBlobEntriesVariables

dehann:
convenience function for gathering many blob entries over variables


Affie (approved):
Looks good, we might just consider having the variable label in the results, but I don't know if your use case needs it.

dehann:
thanks, thought about that a bit and think the varList keyword and dropEmpties gives enough freedom if user wants to keep association of variable to blobs.  case i need it for is just get all the blobs matching the pattern

PR #997, opened 2023-03-27
Packed[Variable/Factor] as a DFGType and nstime as string

Affie:


dehann:
Just linking note as we work through the changes:
- #990

Affie:
I'm going to merge this, so master might be a bit unstable as it will be v0.21 going forward

PR #998, opened 2023-03-29
Aligning blob with sdk

Affie:


src/DataBlobs/services/BlobEntry.jl
```diff
@@ -93,7 +92,9 @@ end
 
 #TODO consider changing this to use `first` as in julia's findfirst - as in findfirstBlobEntry->label::Symbol
 # or getBlobEntryFirst, btw this returns a vector in some other places
-function getBlobEntry(var::AbstractDFGVariable, key::Regex)
+@deprecate getBlobEntry(var::AbstractDFGVariable, key::Regex) getfirstBlobEntry(var, key)
+
+function getfirstBlobEntry(var::AbstractDFGVariable, key::Regex)
```
dehann, line 97:
```suggestion
function getBlobEntryFirst(var::AbstractDFGVariable, key::Regex)
```

src/DataBlobs/services/BlobStores.jl
```diff
@@ -94,21 +98,46 @@ function getBlob(dfg::AbstractDFG, entry::BlobEntry)
     end
     throw(
         KeyError(
-            "could not find $(entry.label), uuid $(entry.id)) in any of the listed blobstores:\n $([s->getKey(s) for (s,v) in stores]))"
+            "could not find $(entry.label), uuid $blobId) in any of the listed blobstores:\n $([s->getKey(s) for (s,v) in stores]))"
```
dehann, line 101:
```suggestion
            "could not find $(entry.label), uuid $(entry.blobId) in any of the listed blobstores:\n $([s->getKey(s) for (s,v) in stores]))"
```

src/DataBlobs/services/BlobStores.jl
```diff
@@ -94,21 +98,46 @@ function getBlob(dfg::AbstractDFG, entry::BlobEntry)
...
         )
     )
 end
 
-addBlob!(dfg::AbstractDFG, entry::BlobEntry, data::T) where T =
-        addBlob!(getBlobStore(dfg, entry.blobstore), entry, data)
+getBlob(store::AbstractBlobStore, entry::BlobEntry) =
+    getBlob(store, entry.originId)
```
dehann, line 107: not sure if you can assume originId here.  May have to look for both blobId first or fall back on originId second.

src/DataBlobs/services/BlobStores.jl
```diff
@@ -94,21 +98,46 @@ function getBlob(dfg::AbstractDFG, entry::BlobEntry)
...
+
+#add 
+addBlob!(dfg::AbstractDFG, entry::BlobEntry, data) =
+    addBlob!(getBlobStore(dfg, entry.blobstore), entry, data)
+
+addBlob!(store::AbstractBlobStore, entry::BlobEntry, data) =
+    addBlob!(store, entry.originId, data)
+    
+addBlob!(store::AbstractBlobStore, data) =
```
dehann, line 116:
```suggestion
# also creates an originId as uuid4
addBlob!(store::AbstractBlobStore, data) =
```

src/DataBlobs/services/BlobStores.jl
```diff
@@ -94,21 +98,46 @@ function getBlob(dfg::AbstractDFG, entry::BlobEntry)
...
+    addBlob!(store, uuid4(), data)
+
+#fallback as not all blobStores use filename
+addBlob!(store::AbstractBlobStore, blobId::UUID, data, ::String) =
```
dehann, line 120: maybe just clarify blobId vs originId here

src/DataBlobs/services/BlobStores.jl
```diff
@@ -94,21 +98,46 @@ function getBlob(dfg::AbstractDFG, entry::BlobEntry)
...
+addBlob!(store::AbstractBlobStore, blobId::UUID, data, ::String) = 
+    addBlob!(store, blobId, data)
+
+#update
+updateBlob!(dfg::AbstractDFG, entry::BlobEntry, data::T) where {T} =
+    updateBlob!(getBlobStore(dfg, entry.blobstore), entry, data)
 
-updateBlob!(dfg::AbstractDFG, entry::BlobEntry, data::T) where T =
-        updateBlob!(getBlobStore(dfg, entry.blobstore), entry, data)
+updateBlob!(store::AbstractBlobStore, entry::BlobEntry, data) =
```
dehann, line 127: again would use blobId as primary or fall back on originId?

dehann (approved):
few comments and suggestions, but overall looks good to me

src/DataBlobs/services/BlobEntry.jl
Affie: Ok, I'll go for that naming, coding decide what way around

src/DataBlobs/services/BlobStores.jl
Affie: Ok, My first aim was just to get all the originId vs blobId calls to go through one place to fix the places that used entry.id. 
I'm not sure of the use of origin vs blob so will change as you say only use originId if blobId is nothing

src/DataBlobs/services/BlobStores.jl
Affie: this is just the id that is used in the blobstore as it can only have one. Dispatch on entry should sort out what to pass in here

PR #1009, opened 2023-04-17
v0.21.1-rc1

Affie:
New release for [Add Packed Factor Variable and VND Compare](https://github.com/JuliaRobotics/DistributedFactorGraphs.jl/commit/6c5c4ecf61ce787ba841fc32a3f168a2d50b5b6d)


Affie:
Only failing because of upstream IIF compat, so merging

Issue #985, opened 2023-03-12
missing field in DFG v0.20: BlobEntry.size


dehann:
Missing field `BlobEntry.size`:
- https://github.com/JuliaRobotics/DistributedFactorGraphs.jl/blob/3e2eca0f55314c6176bdd8506f1075fb16dd3910/src/DataBlobs/entities/BlobEntry.jl#L10

dehann:
xref:
- https://github.com/JuliaRobotics/DistributedFactorGraphs.jl/issues/939

dehann:
Hi @Affie , if v0.20.2, can we close this?

Affie:
> Hi @Affie , if v0.20.2, can we close this?

Sorry typo

Affie:
This is not in the API, so I'm not sure if/when it should be done.

dehann:
done in v0.21.2 as optional field `::Union{Int, Nothing}`

PR #1011, opened 2023-05-16
add optional timestamp size fields BlobEntry, (un)packBlob, docs

dehann:


src/DataBlobs/services/HelpersDataWrapEntryBlob.jl
```diff
@@ -51,11 +51,14 @@ BlobEntry(
     label::Symbol = entry.label, 
     blobstore::Symbol = entry.blobstore, 
     hash::String = entry.hash,
+    size::Int = entry.size,
```
Affie, line 54: I still don't understand why you want to add the new size field to BlobEntry. I would say it lower priority and you can use meta. https://github.com/JuliaRobotics/DistributedFactorGraphs.jl/issues/985#issuecomment-1513828542

src/services/CompareUtils.jl
```diff
@@ -267,7 +267,7 @@ function compareFactor(A::DFGFactor,
   end
   @debug "compareFactor 4/5" TP
   if !(:varValsAll in skip) && hasfield(typeof(getSolverData(A).fnc), :varValsAll)
-    TP = TP & (skipcompute || compareAll(getSolverData(A).fnc.varValsAll, getSolverData(B).fnc.varValsAll, show=show, skip=skip))
+    TP = TP & (skipcompute || compareAll(getSolverData(A).fnc.varValsAll[], getSolverData(B).fnc.varValsAll[], show=show, skip=skip))
```
Affie, line 270: should this not be done separately? Is it part of the IIF fixes?

src/DataBlobs/services/HelpersDataWrapEntryBlob.jl
dehann: it doesn't hurt to have this in there, and is a different use case from metadata.  Based on what I've seen the past months, Id like .size to be in please.

src/services/CompareUtils.jl
dehann: Good catch thanks, yes will do in separate PR as part of the IIF fixes.

src/DataBlobs/services/HelpersDataWrapEntryBlob.jl
Affie: I don't have a problem with .size if you want it, it's just that the API does not support it, so it will just get lost. 
Also just make sure this doesn't break the SDKs.

PR #1042, opened 2023-08-20
fix tests for JL 1.10

dehann:


Affie:
This is not really fixing the test. It should check for the returned values specifically

PR #1043, opened 2023-08-20
small refactor on loading different types

dehann:


src/FileDFG/services/FileDFG.jl
```diff
@@ -166,53 +166,34 @@ function loadDFG!(
...
     @info "Loaded $(length(variables)) variables"#- $(map(v->v.label, variables))"
     @info "Inserting variables into graph..."
     # Adding variables
     map(v -> addVariable!(dfgLoadInto, v), variables)
 
+    packedfacts = @showprogress 1 "loading factors" map(factorFiles) do factorFile
+        jstr = read("$factorFolder/$factorFile", String)
+        return JSON3.read(jstr, PackedFactor)
+    end
+    # FIXME, still creates type instability on `variables` as either `::Factor` or `::DFGFactor{<:}`
```
Affie, line 190: I'm not worried about the type stability here as ::Factor is the only thing with a concrete type. ::DFGFactor's type cannot be known ahead of time.

PR #1048, opened 2023-09-05
bug fix, getData with UUID

dehann:


Affie:
Maybe just add some tests if you don't mind. The coverage on HelpersDataWrapEntryBlob.j is not great.

PR #1077, opened 2024-06-18
defFactorType macro

Affie:
- [ ] Need to still sort out how pack/unpack/convert can work. 
    - **Design decision** is to work with pack / unpack and not convert
- [x] Rename to defFactor**Type**
- [x] upstream `PackedSamplableBelief`


Affie:
Merging to refactor factors branch

Issue #1121, opened 2025-03-17
Standardize solver data noun for variables and factors


Affie:
Decision/Vote: go with `state` for both variables and factors.

Ticket Acceptance Criteria:
- [x] `data` is hard to use, what is better: ?
- [x] must work for Variables and Factors
- [x] must have singular and plural
  - [x] variable has solveKey superposition, therefore satellite nodes.  Not so for factor. 
- [x] should this indicate mutable or immutable somehow?  Chances are this is definitely mutable
- [x] define why is this different from metadata
  - metadata is used by humans and usually set final before any solving or compute operations are performed.
  - this `data_field` has to do with compute or algorithm mutations that the novice human generally does not mutate. 

Currently we have the following nouns referring to solver data:

`VariableNodeData` - Struct
`PackedVariableNodeData` - Struct
`GenericFunctionNodeData` - Struct
`FunctionNodeData` = `GenericFunctionNodeData`
`data` - field in FactorDFG
`solverData` - field in FactorCompute
`solverDataDict` - field in VariableCompute
`solverData` - field in VariableDFG
`FactorData` = `PackedFunctionNodeData{InferenceType}`
`SolverData` -  Retrieve solver data structure stored in a variable/factor.
`VariableSolverData` - for get/add etc.

`FactorOperationalMemory` is also partially in the mix here.

Related:
The observation function is currently part of the factor solver data, see https://github.com/JuliaRobotics/DistributedFactorGraphs.jl/issues/1119 and https://github.com/JuliaRobotics/DistributedFactorGraphs.jl/issues/992

dehann:
yes should be the same noun.  @Affie , please dictate us a noun.  We'll get to the actual work at some point.

----

ideas
- [x] `state`
  - +2 on call 25/3/26 
- [ ] `context`
- [ ] `recall`
- [ ] `retention`
- [ ] `compute`
- [ ] `nodememory`
- [ ] `persist`

don't use
- `parameters` because that is immutable and generally a present

dehann:
on use of `state` for either variable or factor, I find it useful to think about probabilistic chain rule:

the nav problem boils down to two sides of the same coin (think Bayes rule):
- sensors make measurements given the state of the world
- estimation inverts to find the state given the measurements

So, is a measurement a state?

[...]

PR #1127, opened 2025-04-02
Refactoring Factors: data -> state, observation and cache

### LINKED ISSUE #992: Investigate splitting factor operational part from 'constant' part of factor
Why?
- Simplify serialization, pack/unpack factor only depends on the factor and does not need the full factor graph
- Simplify PackedFunctionNodeData{T}, GenericFunctionNodeData{T} where  T <: Union{<:AbstractPackedFactor, <:AbstractFactor, <:FactorOperationalMemory}

Questions to be answered
If it can be split so that factor solver data only contains the working memory serialization becomes easy as solver data is just ignored and rehydrated when unpacked for use by the solver and there can possibly be no PackedFactor. 
- What fields are always constant?
  - DF: Almost everything in `FactorNodeData` is constant and can move to FactorFDG -- but, need to checking how `CCW` is built upon `FactorCompute` 
- Will it impact more complicated factor types?
  No, this should not touch the current `usrfun` field functionallity.
- Should we build factor data (w.r.t. the compute part FactorOperationalMemory/CCW) lazily from DFG and only "inflate" when used in IIF. That is, move `reconstFactorData` out of DFG?
    - JT I believe so, that way we don't need a fg to unpack a factor and don't need to calculate ccw if never used (eg plotting).
 

'constant' in this case means not changed by the solver

In other words:
Should we move non FactorOperationalMemory to DFGFactor: 
Maybe: fnc, multihypo, nullhypo, inflation ?
that way we split solverData <: FactorOperationalMemory and constants

### LINKED ISSUE #1119: Rename FactorDFG.data / fnctype becomes...
@Affie and I had a long frustrating conversation to confirm what we already know:  That factor data fields need to be updated for DFG 1 milestone.  This ticket is the confluence of many other comments including the most recent:
- #1118 

## Context for larger work beyond this issue:
1. [This issue] Add new field to `FactorDFG.observFnc`, and replace associated old accessors such as `@deprecate getObservationFunction(::FactorDFG) getFactorFunction(::FactorCompute)`.
**TAC**
  - [x] spike to check no adverse effect on SDK.rs/c
  - [ ] add new field to `FactorDFG.observFnc` (side by side with old `.data / .fnctype`)
    - Also see performance discussion (or Summary level) `FactorDFG.observFnc_type::String` #992
  - [ ] implement `getObservationFunction(::FactorDFG)`
  - [ ] deprecate `getFactorType` use `getObservationFunction( DFG or Compute )`
2. Move all other FunctionNodeData  fields to FactorDFG - #992 
3. Update CCW construction when building FactorCompute - #992 

----

Additional links from discussion:
- https://github.com/JuliaRobotics/DistributedFactorGraphs.jl/pull/1120

Affie:
Refactor factorData::GenericFunctionNodeData to state, observation and cache from #992 design.

PR should do:
- [x] partially address (FactorState) #1121
- [x] close #992 
- [x] close #1119
- [x] #1077  
- [x] update news
- [x] update docs

----

xref:
- related on Variables: #565


src/entities/DFGFactor.jl
```diff
@@ -198,117 +266,120 @@ Base.@kwdef struct FactorCompute{T, N} <: AbstractDFGFactor
...
+    """Observation function or measurement for this factor.
+    Accessors: [`getObservation`](@ref)(@ref)"""
+    observation::FT
+    """Describes the current state of the factor. Persisted in serialization.
+    Accessors: [`getFactorState`](@ref)"""
+    state::FactorState
+    """Temporary, non-persistent memory used internally by the solver for intermediate numerical computations and buffers.  
+    `workmem` is lazily allocated and only used during factor operations; it is not serialized or retained after solving.
+    Accessors: [`getWorkmem`](@ref), [`setWorkmem!`](@ref)"""
+    workmem::Base.RefValue{<:FactorOperationalMemory} #TODO easy of use vs. performance as container is abstract in any case.
```
Affie: @dehann, I would like some input on the naming here. I asked copilot and it came up with `Workmem` (working memory). I came up with `Opmem` (operational memory). As you know the only implementation here is `CommonConvWrapper`. 
FactorOperationalMemory can also be renamed to FactorWorkingMemory if it is clearer. 

A confusing name such as `rebuildFactorMetadata!` will then become `rebuildFactorWorkmem` since `Metadata` has another meaning.

Affie: I don't think it is part of the SDKs and likely only for advanced (solver) devs.

dehann: Hmm, I thought we had resolved a name for this but cannot remember what it was.  When we last spoke, issue was confusion with word "cache".  We had something better than opmem.  I think we just had "memory" - adding the adj op/operational is too long.

Think I'm sticking with:

- "memory": operational memory used by solver routines during compute in hot-loops etc for maximizing performance of source code. E.g. storing calculated gradients or hypothesis recipes.  Any non-standard in-memory objects should instead use the "cache" field.
- "cache": Custom in-memory location generally associated with the loading or saving of objects, for holding case-by-case in-memory values that accelerate the compute tasks.  Furthermore, "cache" is available for custom designs of e.g. factors whereby the developer wishes to manually craft cached objects during compute operations - this is in contrast to solver default routines which instead use the "memory" field.  E.g. loading lidar blobs from storage into a factor cache for faster rigid transform registration compute.

Affie: You might be right, I also first thought we discussed this already, but figured it was the very similar #1121. The only other related name that I found was `ComputeMemory`, but I think that was just a placeholder name I used? 

The cache vs memory sounds familiar though, the `FactorOperationalMemory` is basically just a cache of other variables.

I don't think we can get away with just memory as `getMemory` is too ambiguous.

dehann: Strictly speaking we may want to flip "cache" and "memory" from what I said above.  In a Von Neumann computer we have RAM memory for holding variables and L cache for hot-loop compute speed...  that would also require we change "preambleCache" to new signature (think that is happening anyway).

Affie: I think I remember now, it was a discussion on preambleCache verb nouns and we didn't come to a conclusion because it's not in DFG, that's why it wasn't added to verb-noun dict. 

I'm in for changing to a cache related noun. It was also one of the suggestions by copilot:
- **FactorWorkMem** is short, clear, and idiomatic for temporary, operational memory.
- **FactorCache** is also very common for lazily (re)hydrated, non-persistent data.
- **FactorScratch** is a classic term for temporary computation memory.

dehann: So I really like:
- `Cache`: operational memory used by default solver routines during compute in hot-loops etc for maximizing performance of source code. E.g. storing calculated gradients, hypothesis recipes, or ode-solutions.  Good solver will keep this `isbitstype == true`.  Any non-standard in-memory objects should instead use the "scratch" field.
- `Scratch`: Custom in-memory location generally associated with the loading or saving of objects, for holding case-by-case in-memory values that accelerate the compute tasks. "Scratch" is available for custom designs of e.g. factors whereby the developer wishes to manually craft in-place objects to optimize compute operations - e.g. loading lidar blobs from storage into a factor scratch as well as faster rigid transform registration compute techniques.  Up to user to be good developer and eventually get `.scratch` to `isbitstype == true`, but doesn't have to be on day one.   In contrast, solver default routines should instead use the "cache" field.

This means that any time a sampler or residual function runs, the user will have access to the `CalcFactor` object's **mutable memory**:
- `calcfactor.scratch`: here user/developer does their custom work to make the machine go.  Also whatever the user setup on factor load/add.  
- `calcfactor.cache`: in-place solver stored values such as gradients or whatever else happens per default.  

Remember the solver design is going down road of always creating a new `CalcFactor` object for each independent thread.  I.e. `calcfactor.*` only lasts as long as the thread/hot-loop  used to compute with it. 

Then there is also the issue of `parching`, e.g. load a `FactorCompute` from DBDFG or FileDFG such as lidar registration where .laz files representing two point clouds are stored as blobs.  Then the `preambleScratch` function would run once and a user is responsible for something like `scratch.pc1 = getBlob(parchedfactor.pointcloud1blobid,...)`.  We have a first example of this, let me see if I can find... yes all the `ObjectAffordSubcloud` stuff: https://github.com/JuliaRobotics/Caesar.jl/blob/9df2df54d4a1bc271afafd5730d35c95af4c67f5/src/objects/ObjectAffordanceSubcloud.jl#L241

src/Deprecated.jl
```diff
@@ -1,6 +1,14 @@
 ## ================================================================================
 ## Deprecated in v0.27
 ##=================================================================================
+export AbstractFactor
+const AbstractFactor = AbstractFactorObservation
+
+export AbstractPackedFactor
+const AbstractPackedFactor = AbstractPackedFactorObservation
```
Affie, line 8: see https://github.com/JuliaRobotics/DistributedFactorGraphs.jl/issues/1119#issuecomment-2979923789

src/entities/DFGFactor.jl
Affie: So far renamed FactorOperationalMemory and related to FactorSolverCache, promoted to an issue for further discussion and to track implementaion.

PR #1128, opened 2025-04-02
Change GraphPlot -> GraphMakie

Affie:
close #1126

PR #1130, opened 2025-05-05
deps bump ManifoldsBase to include v1

dehann:


Project.toml
Affie, line 1: Next DFG should be v0.26.0 if I’m not mistaken.

dehann: HI @Affie, ~~It should be fine to do a patch release.  Not sure why #1128 is labelled breaking change if there are no changes to exports?~~  Only a few other small PRs merged without breaking changes.

Also, I need a patch fix to restore all the repos for stable so trying to make it work before any real breaking changes like 1127 happen...

dehann: i guess you mean user needs a new package and therefore a minor release.  That's probably better -- I'll just release as v0.26.0 then.

Affie: Yes. See news.md using GraphPlot -> using GraphMakie

Issue #1133, opened 2025-05-21
Replace PPE/MKD with new noun (Homotopy?) + Serialization Design


Affie:
## Criteria

- [ ] Select new noun (HybridBelief or HomotopyBelief or ...?)
- [ ] First implementation to replace PPE and keep door open for replacing MKD thereafter
- [ ] Replace PPE as part of DFG v1 work downstream
- [ ] How to serialize associated tree and manifold points

From the original needs
- Ploting - function calls can convert to required type
- Searching or indexing only applicable to certain types.

## Description

Related names are Homotopy*, MKD, GMM, Mixture, ...?

## Storied History

See discussion below

Affie:
- In old world KDE was not on manifold, but had a balanced tree and could approximate non-parametric products.
- Rewrite was Manifold Kernel Density (MKD), now using `ManellicTree` for on manifold unbalanced tree, using Mahony parallel transport for approximating products. Cleaner rewrite of serialization.
- Parametric Point Estimates `PPE ` was window dressing for non-parametric beliefs, but never really supported multiple modes properly. 
- Working towards hybrid parametric non-parametric belief representations. 

As an example, the current PPE mean/max functionality can be achieved with getters such as `getBelief[Max/Mean]`

Affie:
Maybe also something with a tree such as BeliefTree, ModeTree, or MixtureTree

Affie:
The plan is to deprecate PPEs in the meantime, that way we can always add this after DFGv1 as well.

dehann:
lalala from Slack (by Affie):
- just FYI, here is how Mixtures are serialized (the code still has to move out of factors as it is no longer a factor). The NamedTuple was the problem area that needed custom code.
[https://github.com/JuliaRobotics/IncrementalInference.jl/blob/d98ce5744afa5cf57a7a[…]c6fdfc946c118439567/IncrementalInference/src/Factors/Mixture.jl](https://github.com/JuliaRobotics/IncrementalInference.jl/blob/d98ce5744afa5cf57a7a8c6fdfc946c118439567/IncrementalInference/src/Factors/Mixture.jl#L57-L71)

Issue #1085, opened 2024-07-11
Review verbs for DFG v1.0


Affie:
Verbs here: https://github.com/JuliaRobotics/DistributedFactorGraphs.jl/wiki/Standard-Definitions-of-API-via-verbs-and-set-theory
With regard to:
- Signature
- Return Type
- Error Behaviour

Verbs:
- [ ] list
- [ ] get
- [ ] add
- [ ] update
- [ ] delete

TODO add all verbs for DFG v1.0

Affie:
**`list`**
Synchronously list the labels of an existing (composite) data container/object with optional filters.
    - Filters on label: eg. startswith, contains, endswith, Regex (not SDK)
    - Filter on Solvable (only variables and factors)
    - Filter on Tags (tag in node.tags)
    - Filter on Type (only variables and factors)

- Synchronous
- Example nouns Variables, Factors, VariableNodeData, PPE -> Vector{Symbol} 
- Returns empty list if the container exists but empty
- Throw typed Exeption `KeyError` or custom exeption if the container does not exist.

----

Legacy refs:
- #528

Affie:
`get` returns single unique entry else errors.

Affie:
`delete`
Remove key and data at a given key if it exists. 
- Synchronous
- Example nouns Variables, Factors
- Return number of nodes deleted - #1134
- return 0 or try to if label does not exists, so don't throw typed Exception `LabelNotFoundError` . see #1211

Affie:
On `update` we decided that `merge` is a better fit for the current update behaviour.

update can perhaps take keyword arguments of what to update, for example: `updateVariable(fg, label; solvable = 1)`

As used like this in IIF on `updateVariableSolverData!`:
```julia
updateVariableSolverData!(
  dest,
  var,
  solveKey,
  false,
  [:val; :bw; :infoPerCoord; :solvedCount; :initialized];
  warn_if_absent = false,
)
```
Although, this specific signature is quite confusing and not very clear that `var` is not from `dest` fg. Options here are maybe `merge` `sync` `copyto` or similar.

dehann:
I disagree with keyword specification for what to fields to `update`.  This is taking CRUD too far.

I agree with using `merge` for now and keeping `update` as a reserved verb for later.

Therefore, I would expect DFG v1 usage would generally look something like this:

```julia
v1 = getVariable(srcfg, :x1)
[...]

PR #1145, opened 2025-06-25
implement mergeStates!

Affie:
I went back and forth on `mergeVariableStates!` and comparing it with `updateVariableNodeData!` and settled on following the verb by still providing a container. The same concept can be followed for all satellite levels (such as `Blobentry`)

@dehann would you mind giving your inputs.

### DIFF
--- a/src/services/DFGVariable.jl
+++ b/src/services/DFGVariable.jl
@@ -740,6 +740,17 @@ function mergeStates!(dfg::AbstractDFG, varLabel_state_pairs::Vector{Pair{Symbol
     return sum(cnt)
 end
 
+function mergeStates!(
+    dfg::AbstractDFG,
+    variableLabel::Symbol,
+    states::Vector{<:State},
+)
+    cnt = asyncmap(states) do state
+        return mergeState!(dfg, variableLabel, state)
+    end
+    return sum(cnt)
+end
+
 function copytoState!(
     dfg::AbstractDFG,
     variableLabel::Symbol,
--- a/src/services/DFGVariable.jl
+++ b/src/services/DFGVariable.jl
@@ -733,7 +733,7 @@ function mergeState!(v::VariableCompute, vnd::State)
     return 1
 end
 
-function mergeStates!(dfg::AbstractDFG, varLabel_state_pairs::Vector{Pair{Symbol, VariableState}})
+function mergeStates!(dfg::AbstractDFG, varLabel_state_pairs::Vector{Pair{Symbol, State}})
     cnt = asyncmap(varLabel_state_pairs) do (varLabel, state)
         return mergeState!(dfg, varLabel, state)
     end
--- a/src/services/DFGVariable.jl
+++ b/src/services/DFGVariable.jl
@@ -733,6 +733,13 @@ function mergeState!(v::VariableCompute, vnd::State)
     return 1
 end
 
+function mergeStates!(dfg::AbstractDFG, varLabel_state_pairs::Vector{Pair{Symbol, VariableState}})
+    cnt = asyncmap(varLabel_state_pairs) do (varLabel, state)
+        return mergeState!(dfg, varLabel, state)
+    end
+    return sum(cnt)
+end
+
 function copytoState!(
     dfg::AbstractDFG,
     variableLabel::Symbol,
--- a/src/services/DFGVariable.jl
+++ b/src/services/DFGVariable.jl
@@ -740,11 +740,7 @@ function mergeStates!(dfg::AbstractDFG, varLabel_state_pairs::Vector{Pair{Symbol
     return sum(cnt)
 end
 
-function mergeStates!(
-    dfg::AbstractDFG,
-    variableLabel::Symbol,
-    states::Vector{<:State},
-)
+function mergeStates!(dfg::AbstractDFG, variableLabel::Symbol, states::Vector{<:State})
     cnt = asyncmap(states) do state
         return mergeState!(dfg, variableLabel, state)
     end
--- a/test/testBlocks.jl
+++ b/test/testBlocks.jl
@@ -787,7 +787,7 @@ function VSDTestBlock!(fg, v1)
     # Bulk copy update x0
     @test DFG.copytoState!(fg, v1.label, :default, getState(fg, v1.label, :default)) == 1
 
-    @test DFG.mergeStates!(fg, Vector{Pair{Symbol, VariableState}}([:a=>vnd])) == 1
+    @test DFG.mergeStates!(fg, Vector{Pair{Symbol, State}}([:a=>vnd])) == 1
     
     altVnd = vnd |> deepcopy
     keepVnd = getState(getVariable(fg, :a), :parametric) |> deepcopy
--- a/test/testBlocks.jl
+++ b/test/testBlocks.jl
@@ -789,6 +789,9 @@ function VSDTestBlock!(fg, v1)
 
     @test DFG.mergeStates!(fg, Vector{Pair{Symbol, State}}([:a=>vnd])) == 1
     
+    @test DFG.mergeStates!(fg, [:a=>vnd]) == 1
+    @test DFG.mergeStates!(fg, :a, [vnd]) == 1
+    
     altVnd = vnd |> deepcopy
     keepVnd = getState(getVariable(fg, :a), :parametric) |> deepcopy
 
--- a/test/testBlocks.jl
+++ b/test/testBlocks.jl
@@ -787,6 +787,8 @@ function VSDTestBlock!(fg, v1)
     # Bulk copy update x0
     @test DFG.copytoState!(fg, v1.label, :default, getState(fg, v1.label, :default)) == 1
 
+    @test DFG.mergeStates!(fg, Vector{Pair{Symbol, VariableState}}([:a=>vnd])) == 1
+    
     altVnd = vnd |> deepcopy
     keepVnd = getState(getVariable(fg, :a), :parametric) |> deepcopy
 
--- a/test/testBlocks.jl
+++ b/test/testBlocks.jl
@@ -788,10 +788,8 @@ function VSDTestBlock!(fg, v1)
     @test DFG.copytoState!(fg, v1.label, :default, getState(fg, v1.label, :default)) == 1
 
     @test DFG.mergeStates!(fg, Vector{Pair{Symbol, State}}([:a=>vnd])) == 1
-    
     @test DFG.mergeStates!(fg, [:a=>vnd]) == 1
     @test DFG.mergeStates!(fg, :a, [vnd]) == 1
-    
     altVnd = vnd |> deepcopy
     keepVnd = getState(getVariable(fg, :a), :parametric) |> deepcopy
 


test/testBlocks.jl
```diff
@@ -795,8 +795,8 @@ function VSDTestBlock!(fg, v1)
     @test mergeVariableState!(fg, :a, vnd) == 1
 
     # Bulk copy update x0
-    @test updateVariableSolverData!(fg, [v1], :default) == nothing
-
+    @test DFG.mergeVariableStates!(fg, [:a=>vnd]) == 1
```
dehann: Yes on using verb `merge` and dropping `update`, see:
- #1085

src/services/DFGVariable.jl
```diff
@@ -708,6 +708,27 @@ function mergeVariableState!(dfg::AbstractDFG, variablekey::Symbol, vnd::Variabl
...
+    dfg::AbstractDFG,
+    varLabel_state_pairs::Vector{<:Pair{Symbol, <:VariableState}},
+)
+    cnt = asyncmap(varLabel_state_pairs) do (varLabel, state)
+        return mergeVariableState!(dfg, varLabel, state)
+    end
+    return sum(cnt)
+end
+
+function mergeVariableStates!(
```
dehann: I would have expected the default implementation signature to be 
```julia
function mergeState!(<:Variable, <:VariableState)
  # actual implementation lives here
end

mergeState!(dfg::Factorgraph, lb::Symbol, state::VariableState) = mergeState!(getVariable(fg, lb), state)

mergeStates!(dfg::Factorgraph, ::Vector{<:Pair_}) = # pmap over vector mergeState!(dfg, pair[1], pair[2])

mergeVariableState!(w...;kw...) = mergeState!(w...;kw...) # alias
```

dehann (dismissed):

test/testBlocks.jl
Affie: yes, merge is what we decided. Captured in [wiki](https://github.com/JuliaRobotics/DistributedFactorGraphs.jl/wiki/Standard-Definitions-of-API-via-verbNoun-structure)

src/services/DFGVariable.jl
Affie: `mergeState!`  is not consistent with the rest of the API as we have `getVariableState` and therefore would expect `mergeVariableState!`. We need the `Variable` noun in `get` explicit to avoid ambiguities, to have a type stable implementation in julia, and to help with SDKs without multiple dispatch. It also makes for more readable and maintainable code to be more explicit -- when updating from `getSolverData`, it was a pain to distinguish between Variable and Factor. We can still do the alias, but that would likely be a DFG only function and not across SDKs. I would also not include `[get/add/merge/delete]State[!]` for DFGv1 beta work.

Does the comment `... ::Vector{<:Pair_})` mean you are happy with the new signature, what I wanted to know form this PR?

Affie: Also see similar discussion [here](https://github.com/JuliaRobotics/DistributedFactorGraphs.jl/issues/1146) on possible future expansion of FactorState.

dehann: I'm okay with `::Vector{Pair{Symbol, <:VariableState}}`.  There might be some dynamic dispatch if different Variable states are merged, but that is okay too in Julialand -- AND, we are fixing many of those cases separately with the HomotopyBelief* idea:
- #1133

dehann: EDIT: Should add that the signature with implementation vs alias does not really bother the DFG v1 release -- this is only where the functional code actually resides.  Both `getState` and `getVariableState` are still valid according to the DFG v1 spec.  Similar for `mergeState!` vs `mergeVariableState!`.

----

## Don't loose sight of

> therefore would expect

You expect the longer signature, however, I expect zero duplication is better.  For `getVariableState(::VariableDataType)` "variable" noun twice.

> not consistent with the rest of the API

Again, that is what I was trying to warn about in issue 56.  Obviously the overall philosophy of how to build signatures dominates everything BUT I'm not sure we are on the same page.  Here is the top line description from the verbNoun wiki (which has been my mental picture all along):
```
verbNoun[Noun|Adjective][!](::NounType[,..., defaults=defaults; kwargs=defaults])
```

Shorter is better.  The longer `verbNounNoun(::NounType,...)` is opposite to what I understood we had decided before.  I thought we are building the shortest possible name in Julia's dispatch and then adding the longer signatures as aliases.  For non-dispatch/non-overloading language SDKs we only have the longer signatures (and the core implementations are in Julia only anyway).

Do you have a reference/issue/comment somewhere on changing this default philosophy to longer (w/ noun duplication) rather than shorter names (as per wiki)?

> it was a pain to distinguish between Variable and Factor.

I get you were struggling during refactor, but you have to look at this in hindsight (after DFG v1 has landed).  You struggled because the old code was a mess.  Your struggle was hopefully the last time the verbNoun refac needs to go through such as major cleanup.  

Personally, I do not see a problem with using either: `getState(::Variable_,...)` or `getState(:Factor_,...)` and these would return different types.  That's the key, Julia's ability to return different types.  Julia implementations would live in `getState` (similar to old `getSolver[d]ata`), with aliases such as `getFactorState`.

> `mergeState!` [...vs...] `mergeVariableState!`.

`mergeState!` is pretty natural for me.  Inconsistent with rest of API and struggled during refactor sound like transient reasons, and non-dispatch SDKs are basically just shims (glorified aliases) over GQL.  Is there perhaps another reason for more verbose?

I do remember we discussed choice between `getStateVariable` vs `getVariableState` -- again, I was under the impression the decision from there is to only implement `getState` and then include alias `getVariableState`.  In extreme cases maybe e as far as including a second alias `getStateVariable` (bad example).



> It also makes for more readable and maintainable code to be more explicit

Proper use of multiple dispatch reduces maintenance.  Verbose manual write out of what could have been dispatch increases maintenance load.

Less duplication is more readable for me, because the explicit "Variable" in function name tells me there are cases without variables.  I think we should use multiple dispatch properly and only add the longer aliases to have consistency with other non-dispatch SDK languages.

dehann: Also see new docs intro spawned by this discussion:
- JuliaRobotics/Caesar.jl#1037

Affie: Multiple dispatch can’t sort out the getState ambiguity and type stability in the crud get function, your examples are just the getters where multiple dispatch works. 
The CRUD signiture is:
`getVariableState(fg::AbstractDFG, varLabel::Symbol, stateLabel::Symbol)::VariableState` so there is no duplication of variable type. 
In `mergeVariableState[s]!` there is also no variable type but dispatch can happen on the state type. If changed you sometimes have to use the one pattern and sometimes the other.

Affie: From wiki:
> As a general policy, the decision was made to implement the logic inside the shorthand method definition

Thanks, I didn’t see /remember this statement and just followed the rest of DFG/SDKs for a consistent API. Is this decision written somewhere so i can follow the reasoning. I’m not sure that it took into consideration that the SDK.jl should look more like another DFG diver than a completely separate entity as it was first implemented. 
I would prever the standard CRUD be implemented across the board and the shorthand versions should be discouraged if you would like to swop to different technologies (drivers/SDKs)

I’ll look at the codebase again to see how closely it’s followed. I know in BlobEntries are verbose and listVariables is implemented and not ls. But need to go through all of them. 

edit: technically we can just have `add!`, but that is not the case, so I don't think that hase been the design so far. We briefly discussed this i a call and decided to go with `addVariable!` for dfgv1.

Affie: > Personally, I do not see a problem with using either: getState(::Variable_,...) or getState(:Factor_,...) and these would return different types.

The issue there was: as an example some code called `getSolverData(v)` where v was a factor.  Sure julia knows it, but I didn’t. 

Perhaps the real issue here is that the State noun alone is used for 2 different things.
If it was clear enough we wouldn’t need this discussion.

Affie: I can’t think of better nouns than [Variable/Factor]State. 
It does have the same issue as Blobentry (compounded of 2 nouns) though. 
I kept it camel case but should have maybe also gone with compound nouns such as `Var[iable]state` and `Fac[tor]state`.

Affie: Are the type names the issue then?
https://github.com/JuliaRobotics/DistributedFactorGraphs.jl/blob/dd9f43a3bc0d2ff5616c18f5b679b114d0d8602c/src/entities/DFGVariable.jl#L22
https://github.com/JuliaRobotics/DistributedFactorGraphs.jl/blob/dd9f43a3bc0d2ff5616c18f5b679b114d0d8602c/src/entities/DFGFactor.jl#L22

Affie: If you don't like the long names, I would be happy with:
```julia
@kwdef mutable struct Varstate{T <: VarstateType, P, N}
# and 
@kwdef mutable struct Facstate

# CRUD and getters then become:
`getVarstate` and `getFacstate` with no ambiguity.
```
I don't mind verbose names as you never type it out with autocomplete, the shorter names still start with `Var` and `Fac`, therefore autocomplete will still get you to the correct function.

Affie: Just also referencing the original issue here: https://github.com/JuliaRobotics/DistributedFactorGraphs.jl/issues/1121

Affie: Note: `getVariableState` was `getVariableSolverData` in the CRUD get, but I'm open if we can improve it more.

Affie: Related to this, we should also look into what to do about the plural variable versions that SDK.rs can use to avoid function overloading - such as (being overly verbose for clarity):

1.
```julia
function mergeVariablesVariablestates!(
    dfg::AbstractDFG,
    varLabel_state_pairs::Vector{<:Pair{Symbol, <:VariableState}},
)
```
2.
```julia
function mergeVariableVariablestates!(
    dfg::AbstractDFG,
    variableLabel::Symbol,
    states::Vector{<:VariableState},
)
```
I'm not sure if rust wil be able to do the function overloading in some way if we just have (no plural on variable): `mergeVariableVariablestates!`/`mergeVariablestates!`/`mergeVarstates!`. (depending on the outcome above).

I'm leaning towards just `mergeVar[iable]states!`, but will have to see what can be used consistently across the API.

Affie: copilot's opinion:
Between Variablestate and Varstate, both are clear, but here’s a comparison based on your API style and the conventions in your code:

Varstate

Pros: Concise, easy to type, and fits well with your noun-based, single-capital convention.
Cons: Slightly less explicit, but still clear in context.
Variablestate

Pros: More explicit, immediately clear to new users what it refers to.
Cons: Longer, and may feel redundant since "variable" is already implied by context.
Recommendation:
If your API values brevity and you document your naming conventions, Varstate is the better choice.
If you want maximum explicitness for new users, Variablestate is also fine, but may be unnecessarily verbose.

Summary:
Varstate is preferred for your style—concise, clear, and consistent with your API.

Affie: Also just cross referencing https://github.com/NavAbility/NavAbilitySDK.jl/pull/288, where `getVariableState` was approved as API. We should change that also pending this outcome.

Affie:
Note, branch was rebased to resolve conflics.

Affie:
Just noting differentiating between the accessor-like `get` and the SDK-CRUD `get`:
- CRUD signatures that has a `AbstractDFG`, `AbstractBlobstore`, etc. as first argument should be implemented across the SDKs and save to use.

src/services/DFGVariable.jl
Affie: `exists` is another similar case. Variables and factors use just `exists` and all other nouns use `hasObj`. (The verb noun wiki says we are going with `has`, so just giving as an example)
The signature is just `exists(fg, label)` so we need to read the documentation that it will look for variable or factor existence where hasVaraible and hasFactor is a more self explanatory.  What about [graph|agent]blobentries/metadata, etc. existence? With has... autocomplete gets you there. We can consider keeping `exists` as a DFG function that calls to `has` internally. That is not "implement the logic inside the shorthand method definition", but I don't really see the value in that as a hard rule and more implementation specific.

dehann: > edit: technically we can just have add!, but that is not the case, so I don't think that hase been the design so far. We briefly discussed this i a call and decided to go with addVariable! for dfgv1.

I remember from this discussion that logic would go in the shorthand and alias for the longer signatures to match other languages that cannot dispatch.

> Do you have a reference/issue/comment somewhere on changing this default philosophy to longer (w/ noun duplication) rather than shorter names (as per wiki)?
> Is this decision written somewhere so i can follow the reasoning.

yep, I'm asking the same question.

> Personally, I do not see a problem with using either: getState(::Variable_,...) or getState(:Factor_,...) and these would return different types. [skipping your reply as vague but will continue from the phone call conversation that was better]

I previously missed your point that dispatch, although near infinite, is not always type stable.  So `getState(::AbstractDFG, ::Symbol)` could return either `FactorState` or `VariableState` from the same signature -- hence require slow dynamic dispatch.  This for me is a good reason for a more verbose signature.  HOWEVER, the shorter signature for `getState(::AbstractVariable / ::AbstractFactor)` should still produce type stable code and therefore good.  Only the highest level API calls may be a problem for short signatures, but lower down should be good.

> Also just cross referencing https://github.com/NavAbility/NavAbilitySDK.jl/pull/288, where getVariableState was approved as API. We should change that also pending this outcome.

From phone call, the answer is likely DFG v1.x where shorter signatures are added or just logic moved around, but no breaking changes.  Sounds like most important is just getting the verb nouns consistent all over as necessary step -- i.e. focus on that for now.

> `mergeVariablesVariablestates!`

This is exactly why I favor the short names with dispatch and don't want to handicap Julia dispatch design for the sake of the other languages.  So I would add the verbose signatures as aliases over short dispatches like just `add!` or `merge!` -- essence of good design once we get there (will take time to get that clean a design though).

For this case I would just go with `mergeVariableStates!`:
```julia
function mergeVariableStates!(
    ::AbstractDFG,
    ::Symbol,
    ::Vector{<:Pair{Symbol, <:VariableState}},
)
```

> `getVarstate` and `getFacstate`

I don't like these short names.  In that case better to go with longer `getVariableState` or `getFactorState`.

> copilot's opinion:

Nope, I disagree with copilot but thanks for checking and copying that over.

dehann: > exists is another similar case. Variables and factors use just exists and all other nouns use hasObj. (The verb noun wiki says we are going with has, so just giving as an example)

Sounds like the verbNoun wiki/dictionary is wrong.  While `has` and `exist` may both be there, the definitions should make it clear which one is the best.

> We can consider keeping exists as a DFG function that calls to has internally.

depends on what the best definitions for `exist` and `has` are.  If the definitions are bad then I suggest you fix/change/update the definitions first and then update the code second.  Don't just follow the dictionary if it doesn't make sense, change the dictionary.

> The signature is just exists(fg, label) so we need to read the documentation that it will look for variable or factor. What about [graph|agent]blobentries/metadata, etc. existence? With has... autocomplete gets you there.

Again, I would expect to just find `exists` and tab complete for 100's of dispatches with a single clear docustring, and a few related examples (but limited to stable types per signature).  Pretty sure that is what core Julia is doing already.  I'd be okay to compromise for now with always searching both variable and factors and rather preserve the short signature `exists(fg, lbl)` , while avoiding dynamic types..

> We can consider keeping exists as a DFG function that calls to has internally. That is not "implement the logic inside the shorthand method definition", but I don't really see the value in that as a hard rule and more implementation specific.

on this I would again just say that when the design is perfect, then only a few high level dispatches would cover everything, but that quality design is going to take time and instead solve under DFG v1.x guidelines -- i.e. make the decision that avoids breaking changes during DFG v1.x series releases and that allows the signatures to become shorter and more Julia dispatchy as the minor releases roll.

dehann (commented):
see topic by topic responses

src/services/DFGVariable.jl
Affie: >>mergeVariablesVariablestates!

> This is exactly why I favor the short names with dispatch and don't want to handicap Julia dispatch design for the sake of the other languages. So I would add the verbose signatures as aliases over short dispatches like just add! or merge! -- essence of good design once we get there (will take time to get that clean a design though).
>I don't like these short names. In that case better to go with longer getVariableState or getFactorState.
and other replies...

I think you are still missing the issue here. The current nouns are `VariableState` and `FactorState` not just `State`
- that breaks the capital rule as with Blob**E**ntry vs Blob**e**ntry -- but this as a rule might be a mistake
- That causes the shorthand to be `getVariableState` and the long version to be `getVariableVariableState`

So, let me ask directly in another way, how would you like to distinguish between `VariableState` and `FactorState` structs?
Keep in mind that that will be the Noun used in CRUD.

Affie: From triage call: 
@dehann suggested `Recipe` as an alternate name for `FactorState`. In this case `VariableState` will just become `State`. Therefore, crud will be, for example,  `getState` and `getRecipe`. 
Factors only have one `Recipe` and not keyed by a label.

Reminder, this is the factor data that will become Recipe: 
```julia
mutable struct FactorState
    eliminated::Bool = false
    potentialused::Bool = false
    multihypo::Vector{Float64} = Float64[] # TODO re-evaluate after refactoring w #477
    certainhypo::Vector{Int} = Int[]
    nullhypo::Float64 = 0.0
    solveInProgress::Int = 0 #TODO maybe deprecated or move to operational memory, also why Int?
    inflation::Float64 = 0.0
end
```

Affie: That sounds good to me.
`State` is clear enough on its own especially for someone coming from state estimation. The factor - Recipe is more advanced and used internally, `Observation` is the noun that wil used more externally.

Affie: `FactorState` -> `Recipe` looks ok to change, it doesn't feel perfect yet. It will not be exported; therefore, we don't need to sort it out just yet.  We can finalize the structure with the final serialization structure.

Affie:
I rebased again and changed the base to match new naming from #1163. It is now just `mergeStates!`.

Issue #1153, opened 2025-07-02
Standardize Abstract types


Affie:
The Abstract prefix came up in #1146

Listing current abstract types for "do we use Abstract prefix?":

| AbstractType | JustType |
|---|---|
| **AbstractDFG** | |
| **AbstractParams** |   |
| **AbstractBlobstore** |  |
|  | **DFGNode** | | 
| **AbstractDFGVariable** |  |
| **AbstractDFGFactor** |  |
| **AbstractPackedFactorObservation** | |
| **AbstractFactorObservation**|  |
| **AbstractPrior** |  |
| **AbstractRelative** | |
| | **FactorSolverCache** | 
| | **VariableStateType**  |
| | **PackedBelief** |
| AbstractPointParametricEst | will be obsolete |

Note: DFGVariable and DFGFactor conflicts with names to be deprecated and is too similar to VariableDFG.

Julia uses the Abstract prefix regularly, and it's a recommendation in the SciML style guide: https://docs.sciml.ai/SciMLStyle/stable/#General-Naming-Principles

---

### DFG v1.0 implementation :
| Old                              | AbstractType                | Alias               |
|----------------------------------|-----------------------------|---------------------|
|                    | AbstractDFG                 | --                  |
| _AbstractParams_                 | AbstractDFGParams           | DFGParams           |
|               | AbstractBlobstore           | Blobstore           |
| _DFGNode_                        | AbstractGraphNode           | GraphNode           |
| _AbstractDFGVariable_            | AbstractGraphVariable       | GraphVariable       |
| _AbstractDFGFactor_              | AbstractGraphFactor         | GraphFactor         |
| _AbstractPackedFactorObservation_| AbstractPackedObservation   | PackedObservation   |
| _AbstractFactorObservation_      | AbstractObservation         | Observation         |
| _AbstractPrior_                  | AbstractPriorObservation    | PriorObservation    |
| _AbstractRelative_               | AbstractRelativeObservation | RelativeObservation |
| _FactorSolverCache_              | AbstractFactorCache         | FactorCache         |
| _VariableStateType_              | AbstractStateType           | StateType           |
|                    | AbstractPackedBelief        | PackedBelief        |

Affie:
We can maybe just use Abstract in definitions and then alias:
```julia
julia> abstract type AbstractFoo end
julia> const Foo = AbstractFoo
AbstractFoo
julia> Foo
AbstractFoo
```

[...]

Issue #1146, opened 2025-06-25
Review abstract types and names used in IIF that is exported from DFG


Affie:
To review
- [x] InferenceVariable
- [x] AbstractRelativeMinimize
- [x] AbstractManifoldMinimize
- [x] pack/unpackDistribution
- [x] SamplableBelief
- [x] AbstractPrior -> PriorObservation
- [x] AbstractRelative -> RelativeObservation
- TODO fill in more..

other
- InferenceType used in SDK should be removed

Affie:
1. I'm happy with `AbstractPrior` and `AbstractRelative`, but we can add Observation if need for clarity eg. `AbstractRelativeObservation`

2. The noun [pack/unpack]Distribution and SamplableBelief refers to the same thing as far as I can tell, so should we combine them? `Distribution` in SDK is currently a duplicate of `PackedSamplableBelief` and should be consolidated and deprecated.

3. `AbstractRelativeMinimize` and `AbstractManifoldMinimize` should maybe become traits to be expandable, eg. `getSolverMethod(::Pose2Pose2) = ManifoldMinimization()` with the default to AbstractManifoldMinimize - I think we are moving to only one method in any case? But traits open us up to change easier in the future. 

4. Is `InferenceVariable` clear enough for new users (used in declaring variables eg. `struct Position{N} <: InferenceVariable end`) and does it match with the new `State` noun?  `InferenceType` in the SDK refers to the factor type adding to the confusion. `VariableType` is used in the SDK for the same thing. Variable type is also confusing as we have a few variable types (VariableCompute, VariableDFG, etc.) . Maybe `[Abstract]VariableStateType`?

5. There is currently a mix of `InferenceVariable`s with a dimension parameter and without, we should maybe just formalize to always have a parameter (or never as it might just be `Position{N}`), eg. Pose2 becomes Pose{2} with alias as well. `InferenceVariable` would become `InferenceVariable{N<:Integer}` where N is always the dimension.

dehann:
> I'm happy with AbstractPrior and AbstractRelative, but we can add Observation if need for clarity eg. AbstractRelativeObservation

+1 for adding `Observation`, but why not just use `RelativeObservation` and `PriorObservation`.  I feel the abstract designator is already available, so don't really need to repeat it in the signature -- i.e. `isabstracttype(RelativeObservation) --> true`.

[...]

PR #1154, opened 2025-07-03
Rename IIF abstract types

Affie:
Implements #1146
Not the final names, just a little more standardized. Final naming DBD from #1153


Affie:
Test passed against IIF branch

PR #1220, opened 2026-04-16
Rebuild the Blob system as a CAS model

### LINKED ISSUE #1157: Make blobId=>Blob immutable
According to [verb noun wiki](https://github.com/JuliaRobotics/DistributedFactorGraphs.jl/wiki/Standard-Definitions-of-API-via-verbNoun-structure#noun-object), `Blob` is immutable. Currently, that is not enforceable because you can just delete the blob and then add a new blob with the same blobId. I would still like to be able to delete blobs.  Therefore, the pair blobId=>blob should become immutable #1160.  Even if we make it immutable within a single blob store, it will still depend on the user to maintain the integrity across blob stores, perhaps using uuid7 can also help.

Immutability is also impossible to guarantee in the InMemoryBlobStore, so is the immutable Blob a best effort requirement or should DFGv1 not dictate blob Immutability and just provide all (or just some of) the methods.

Affie:
This PR rebuilds the Blob subsystem into a Content-Addressable Storage (CAS) model keyed by `Multihash`, replacing UUID-based Blobstores with routable Blobproviders.

**Changes:**
- Introduces `Multihash` and a new Layer-1 Blobprovider API (`putBlob!`, `fetchBlob`, `purgeBlob!`) with provider mounting on DFGs.
- Refactors `Blobentry` to store `multihash` + `provider` hint, updates Layer-3 save/load wrappers, and adds compatibility aliases/deprecations.
- Updates serialization, docs, and the full test suite to the new CAS/provider model; removes CSV-based LinkStore/RowBlobstore code and the CSV dependency.

close #1157


src/Blobproviders/Blobproviders.jl
```diff
@@ -0,0 +1,257 @@
...
+        return Multihash(hash_func, io)
+    end
+    return putBlob!(provider, m, filepath)
+end
+
+function putBlob!(provider::LinkBlobprovider, m::Multihash, filepath::String)
+    target_path = blobfilename(provider, m)
+    if !isfile(target_path)
+        mkpath(dirname(target_path))
+        hardlink(filepath, target_path)
```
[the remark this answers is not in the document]
Affie: This store is for links and should not implicitly copy

src/entities/Blobentry.jl
```diff
@@ -225,30 +218,39 @@ hasBlobentry(node, label::Symbol) = haskey(refBlobentries(node), label)
...
+- Returns `nothing` if only the multihash is present but the algorithm is unregistered.
 """
 function checkHash(entry::Blobentry, blob)
+    # Reverse lookup: multicodec code -> hash function
+    code_to_func = Dict{UInt64, Function}(v => k for (k, v) in MULTIHASH_FUNCTIONS)
+
+    code, stored_digest = decode(entry.multihash)
+    func = get(code_to_func, code, nothing)
+    if isnothing(func)
+        @warn "checkHash: unregistered multihash algorithm code $(repr(code)), skipping multihash check"
```
Affie:
```suggestion
        @warn "checkHash: unregistered multihash algorithm code $(repr(code)), skipping multihash check"
        return nothing
```

src/entities/Blobentry.jl
```diff
@@ -225,30 +218,39 @@ hasBlobentry(node, label::Symbol) = haskey(refBlobentries(node), label)
...
 
-- Returns `true` if all present hashes (`crchash`, `shahash`) match the computed values from `blob`.
-- Returns `false` if any present hash does not match.
-- Returns `nothing` if no hashes are stored in the `Blobentry` to check against.
+- Verifies the `multihash` by decoding its algorithm code, recomputing the digest
+  from `blob`, and comparing it to the stored digest.
+- Additionally verifies the `crchash` (crc32c) if present.
+- Returns `true` if all present hashes match.
+- Returns `false` if any hash does not match.
+- Returns `nothing` if only the multihash is present but the algorithm is unregistered.
```
Affie:
```suggestion
- Returns `nothing` if  the multihash algorithm is unregistered.
```

src/services/blobprovider_ops.jl
```diff
@@ -0,0 +1,185 @@
...
+    m = Multihash(hash_func, blob)
+    return putBlob!(provider, m, blob)
+end
+
+#has
+function hasBlob(provider::AbstractBlobprovider, entry::Blobentry)
+    return hasBlob(provider, entry.multihash)
+end
+function hasBlob(dfg::AbstractDFG, entry::Blobentry)
+    # CAS: check all providers — the blob may have been stored via a different route
```
Affie, line 173:
```suggestion
    # CAS: check all providers — the blob may have been stored via a different route
    #TODO check entry.provider first
```

src/services/blobstore_ops.jl
```diff
@@ -1,171 +1,184 @@
...
-#     destEntries = listBlobs(destStore)
-#     typeof(sourceEntries) != typeof(destEntries) && error("Can't merge stores, source has entries of type $(typeof(sourceEntries)), destination has entries of type $(typeof(destEntries)).")
-#     # Same source/destination check
-#     sourceStore == destStore && error("Can't specify same store for source and destination.")
-#     # Otherwise, continue
-#     for sourceEntry in sourceEntries
-#         addBlob!(destStore, deepcopy(sourceEntry), getBlob(sourceStore, sourceEntry))
-#     end
-#     return sourceEntries
-# end
```
Affie, line 171: Open an issue to remember to implement a function to `sync` blobs between blob providers. We now have a flat structure so the steps would be:
- gather list of blobs (multihash ids) to move using Blobentries. (list where...)
- fetch |>  put by ids (maybe add a function for this)

PR #1231, opened 2026-05-07
redo HomotopyDensityDFG, creep change serialization approach

### LINKED ISSUE #535: Don
i think we should solve @GearsAD ‘s concerns regarding softtype on CG first before we close this?  I’m happy with name only if it solves the right problem.  I’m worried about the CG stuff going down a Gaussian-only road and end up forcing the overall DFG side to a crisp but overly restrictive DataLevel1 space.  I mean that users just want Level1 and moan it’s not working well enough, and somehow get out of using Level2.  To help avoid all that i think softtypename::Symbol/String is good.

Just to reiterate, i originally wanted to avoid duplication of the whole softtype so just guessed name as a string as cheap hack.  Without the necessary type information for non-Julia use cases, i’m not sure... Again the question for me is what do we want from Summary (what is the one sentence existence proof for Summary/Level1?).

Is it fair for a user to avoid the Julia internals just to get the inference product in a usable way, or should they at least don some of the Julia-ness to use the inference products?  Perhaps i should add, (again) that we are not building Gaussian-only, so the user should get the full beliefs, not just hide behind PPE.  That means DataLevel2.  I don’t want to dumb down the output and look shoddy against a more mature Gaussian-only system.  I’d rather force the user into Julia or build a better multilang interface (I vote the latter). 

I know Neo4j has a bunch of nice features for Gaussian-only, but that should be “supportive” not “authoritative” in the design.

I would argue Summary/Level1 CANNOT be it is the final inference product.  The final inference product is and always should be DataLevel2.

TL;DR I fairly strongly feel softtypename as Symbol/String or remove completely from Level1.  If there are cool features in Neo4j we can leverage, then leave that door open;  but the DB must never dictate our design for what we want to store, DB must store what is needed for the problem — problem here is *strongly* non-Gaussian (ie not just PPE).

_Originally posted by @dehann in https://github.com/JuliaRobotics/DistributedFactorGraphs.jl/issues/237#issuecomment-655220046_

dehann:
Alternative redo of -- creep change implementation of serialization:
- #1230

At the time of writing, 1230 is good enough for AMP  v0.15.0-alpha tests  to pass.  This redo is easiest path to make sure DFG serialization also works for new HomotopyDensityDFG (replaces StoredHomotopyBelief).  Some detail on abstracts in new HomotopyRepr


Affie:
@dehann, i already took this PR and cleaned up quite a bit so we must just re-coordinate then. I removed the duplicate traits.

src/entities/State.jl
```diff
@@ -1,164 +1,3 @@
...
-
-    # --- Topology (The Hierarchy) ---
-    """
-    L1 Internal Topology: mean_parents[i] = j means means[i] is a child of means[j]. A value of 0 indicates a Root node.
-    """
-    mean_parents::Vector{Int} = Int[]
-    """
-    L2-to-L1 Bridge: point_parents[i] = j means points[i] is governed by means[j]. Points are leaves.
-    """
-    point_parents::Vector{Int} = Int[]
```
Affie, line 113: This was a really efficient design for serialization compared to the sparse array approach. Just writing this comment as a reminder to come back to it.

src/Deprecated.jl
```diff
@@ -761,3 +761,40 @@ end
...
+# refCovariances(state) = refPrincipalForms(state)       # RootsOnly view  
+# refBandwidth(state) = refTrailingForms(state)[1]       # LeavesOnly view
+# But my (JT) preference is for HomotopyBeliefDFG to contain a neutral homotopy tree of nodes
+# and use tree accessors on the node level and not raw references. 
+
+refMeans(state::State) = state.belief.principal_elements
+refCovariances(state::State) = state.belief.principal_forms
+refWeights(state::State) = state.belief.weights
+refPoints(state::State) = state.belief.points
+refBandwidth(state::State) = state.belief.trailing_forms[1]
```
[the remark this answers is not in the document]
Affie: refBandwidth is a transition function and just for the use case of storing the first bw.

test/testSerializingVariables.jl
```diff
@@ -46,27 +46,25 @@ function make_test_variable()
...
         G = DFG.getManifold(Pose{3}())
         push!(bel.points, rand(G, ArrayPartition))
 
         jstr = JSON.json(bel; pretty = true, style = DFG.DFGJSONStyle())
-        parsed = JSON.parse(jstr, DFG.StoredHomotopyBelief; style = DFG.DFGJSONStyle())
+        parsed = JSON.parse(jstr, DFG.HomotopyDensityDFG; style = DFG.DFGJSONStyle())
         @test bel == parsed
     end
 
     @testset "RootsOnlyTopology round-trip" begin
```
Affie:
```suggestion
    @testset "HomotopyDensityDFG round-trip with roots only" begin
```

PR #1255, opened 2026-07-11
solved or count if no statelabel

dehann:


src/entities/HomotopyDensity.jl
```diff
@@ -37,8 +37,9 @@ abstract type AbstractPartialTrait end
     topologykind::HT
     formkind::RF
     statekind::T
-    partial::PT = nothing & (ignore = true,) #TODO deprecated field, still needed for AMP.  
-    # partials field needed for AMP v0.15, will be JSON ignored in DFG v0.29, needs better solution by DFG v0.30
+    # FIXME, rename to subbmersion or deprecate completely?
+    """ partials field needed for AMP v0.15, will be JSON ignored in DFG v0.29, needs better solution by DFG v0.30 """
+    partial::PT = nothing & (ignore = true,)
```
Affie, line 42: The submersion is a map and has to be explicitly provided. This example of 2 `e1 = [1, 0, 0]` "partials" shows why:
- `f(R) = vee(log(R))ᵀ * e₁  # SO(3) -> ℝ¹`
- `f(R) = R * e₁             # SO(3) -> S²`

src/services/variable_ops.jl
```diff
@@ -211,7 +211,11 @@ isSolved, setSolvedCount!
 """
 getSolvedCount(v::State) = v.solves
 function getSolvedCount(v::VariableDFG, solveKey::Symbol = :default)
-    return getState(v, solveKey) |> getSolvedCount
+    if hasState(v, solveKey)
+        return getState(v, solveKey) |> getSolvedCount
+    else
+        return 0
+    end
 end
```
Affie, line 219: Note: All functions with deprecated `solveKey` name is not DFG v1 ready yet and un-exported. 

Looking at this again, the verb and noun are both incorrect.
- the noun should match the field name `solves` 
- for the verb we discussed `tally`: _Calculates value of countable field (ex. number of solves completed)_

Affie: `tallySolves(v::VariableDFG)` should still count the solves across all states though, so no default state_label.

src/services/variable_ops.jl
```diff
@@ -250,7 +254,11 @@ getSolved, setSolved!
 """
 isSolved(v::State) = 0 < v.solves
 function isSolved(v::VariableDFG, solveKey::Symbol = :default)
-    return getState(v, solveKey) |> isSolved
+    if hasState(v, solveKey)
+        return getState(v, solveKey) |> isSolved
+    else
+        return false
+    end
```
Affie, line 261: I would use the direct field access here as discussed (TODO  find ref).

For a `list` alternative we have `listStates` and can add a `whereSolves` filter, ie.  `whereSolves = >(0)`.

